PR #4, opened 2016-09-20
Service worker

mattclough1:


sw.js
```diff
@@ -0,0 +1,120 @@
...
+// Send message to socket connection
+function sendToSocket(rpc) {
+    socket.send(rpc);
+}
+
+function disconnectFromSocket() {
+    if (socket) {
+        if(socket.readyState === socket.OPEN) {
+            socket.onclose = function () {
+                socket.close();
```
danielfarrell: Does this actually close the socket?  Looks more like it sets up an onclose handler.

PR #37, opened 2017-08-22
Add System Capabilities

JackLivio:
Adds navigation and phone capabilities to the ui.getCapabilitiesResponse.


JackLivio:
Implements #36

PR #48, opened 2017-11-01
Fix on hmi status

JackLivio:
Fixes logic for handling perform interaction and alert to make sure hmi status is always properly returned to mobile device


jacobkeeler:
Approving since this would be out of the scope of this PR.

Issue #54, opened 2017-12-05
In App Menu Issues


JackLivio:
There is no back button after the user selects a menu, so they must either press a menu item or return to the app list. 

Also after making a menu item selection, the app is navigated to the MEDIA template regardless of previously set display layouts.

PR #55, opened 2017-12-05
In App Menu Fixes

JackLivio:
App will return to appropriate template after user makes a command selection. Also the user has the option to not make any menu selection and return back to the app view by pressing the menu button again.


JackLivio:
fixes #54

PR #64, opened 2018-03-19
Add TextButtonsOnly Template

JackLivio:


jacobkeeler:
My only concern is that nothing appears if you send more than 6 softbuttons in a single `Show`, but a `SUCCESS` message is received

jacobkeeler:
Also, sending an empty softbutton list does not clear out the existing softbuttons.

JackLivio:
Thank you for the review I will look into these fixes.

PR #65, opened 2018-03-19
 TextButtons With Graphic Template

JackLivio:


jacobkeeler:
Same issue I had with #64, no buttons appear if you send more than 6 softbuttons.

PR #66, opened 2018-03-19
Add Template - GraphicWithTextButtons

JackLivio:


jacobkeeler:
Same comment as #64 and #65

PR #95, opened 2018-06-13
Feature/speak tts audio file

JackLivio:
Implements text to speech for audio files and text ttsChunk types.


src/js/Controllers/TTSController.js
```diff
@@ -18,6 +88,20 @@ class TTSController {
...
+                }
+
+                if(this.filePlaylist.length > 0) {
+                    if(this.filePlaylist[0].type === "FILE") {
+                        this.playAudio();
+                    } else if (this.filePlaylist[0].type === "TEXT"){
+                        this.speak();
+                    }
+                }
+
```
jacobkeeler: Should also be added to these (assuming they are implemented):
1. `Alert` 
2. `PerformAudioPassThru`
3. `AlertManeuver`

src/js/Controllers/TTSController.js
```diff
@@ -2,10 +2,80 @@ import RpcFactory from './RpcFactory'
...
+        this.audioPlayer.onended = () => {
+            this.audioPlayer.src ="";
+            if(this.filePlaylist[0]) {
+                if(this.filePlaylist[0].type === "FILE") {
+                    this.playAudio();
+                } else if (this.filePlaylist[0].type === "TEXT"){
+                    this.speak();
+                }
+            }
+        }
```
jacobkeeler: Should have an error function for this as well, sometimes the audio file type is not supported.

src/js/Controllers/TTSController.js
JackLivio: These are covered by TTS.SPEAK on the HMI side

src/js/Controllers/TTSController.js
JackLivio: updated

PR #103, opened 2018-07-19
Feature/audio streaming indicator

JackLivio:
Adds logic to accept audio streaming indicator to change the play/pause button.

[Related proposal](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0109-set-audio-streaming-indicator.md)


src/js/containers/Buttons.js
```diff
@@ -42,7 +42,7 @@ const mapStateToProps = (state) => {
             icon: iconSeekLeft
         })
     }
-    if (subscribedButtons.OK === true && app.updateMode === "PAUSE") {
+    if (subscribedButtons.OK === true && (app.audioStreamingIndicator === "PLAY" || app.updateMode === "PAUSE")) {
```
jacobkeeler: audioStreamingIndicator should take precedence if it is present, updateMode should only be used if it isn't set or is `PLAY_PAUSE`.

PR #120, opened 2018-10-17
Check for requestType Proprietary

JackLivio:
Handle SystemRequest RequestType. This prevents the hmi from sending other types of systemRequest to be processed by the external policy manager.


src/js/Controllers/BCController.js
```diff
@@ -50,6 +50,10 @@ class BCController {
                 sdlController.getURLS(7)
                 return true;
             case "SystemRequest":
+                if (rpc.params.requestType != "PROPRIETARY") {
+                    // Generic HMI can only process PROPREITARY System Requests
```
jacobkeeler:
```suggestion
                    // Generic HMI can only process PROPRIETARY System Requests
```

PR #125, opened 2018-10-22
Respond to Capabilities

JackLivio:
TTS, UI, and Button controllers will respond to the GetCapabilities Requests from core at startup.


src/js/Controllers/DisplayCapabilities.js
```diff
@@ -2123,6 +2123,17 @@ let capabilities = {
...
+	"COMMON": {
+		"audioPassThruCapabilities": {
+			"samplingRate": "44KHZ",
+			"bitsPerSample": "RATE_8_BIT",
+			"audioType": "PCM"
+		},
+		"speechCapabilities": [
+			"FILE"
+		],
+		"prerecordedSpeechCapabilities": ["INITIAL_JINGLE"]
```
jacobkeeler: `hmiCapabilities` and `systemCapabilities` should be here as well

PR #126, opened 2018-10-23
Add Nav Controller

Jack-Byrne:
Handle navigation video streaming related RPCs.


src/js/Controllers/NavController.js
```diff
@@ -0,0 +1,24 @@
...
+    }
+    handleRPC(rpc) {
+        let methodName = rpc.method.split(".")[1]
+        switch (methodName) {
+            case "StartStream":
+                var message = "This system does not support video streaming."
+                return {"rpc": RpcFactory.UnsupportedResourceResponse(rpc, message)};
+            case "SetVideoConfig":
+                var message = "This system does not support video streaming."
+                return {"rpc": RpcFactory.UnsupportedResourceResponse(rpc, message)};
```
jacobkeeler: Shouldn't there be a default case? Otherwise, the HMI won't send any response to other Navigation RPCs.

jacobkeeler: Should probably directly respond to `IsReady` with `available=false` as well

jacobkeeler: Should also respond to Audio Streaming RPCs.

PR #144, opened 2018-12-10
Tiles only image fix

JackLivio:
Move softbutton image above text. Increase image size.


iCollin:
Closed in favor of #210

PR #167, opened 2019-05-29
App Service Consumer Display

JackLivio:
Adds two new views for displaying app services. Publish an AppService and send OnAppServiceData in order to see available info.

Implemented according to the 5.1 spec for Media, Weather, and Navigation service types.

[Invision App Demo](https://invis.io/WXRYR8MSVDN)


src/js/AppServicesNav.js
```diff
@@ -0,0 +1,238 @@
...
+                        (weatherActive ? "" : " hide-tab")}
+                        onClick={() => this.setActive("weather")}>
+                        <WeatherButton temp={weatherData.currentTemperature} weatherIcon={weatherData.weatherIcon}/>
+                        <div className={"t-small t-medium th-f-color tab-text even-spacing " + ((activeServiceTab == "weather") ? "" : "hide-tab")}>
+                            <div className="weather-item">
+                                <p>
+                                    {weatherData.high ? weatherData.high + "°" : null}
+                                </p>
+                                <p className="t-light">
+                                    {weatherData.low ? "/ " + weatherData.low + "°" : null}
```
jacobkeeler: This looks rather odd if no `temperatureHigh` is provided.

jacobkeeler:
Other minor details:
1. Maybe include placeholder text if no OnAppServiceData has been received, currently the space is just left blank
2. Maybe include an icon to make it clearer that the side button is used to open the full app service card view
3. Maybe include an option to disable animations in the nav bar, since I could see it looking extra choppy on a slower machine

src/js/AppServicesNav.js
```diff
@@ -173,23 +194,26 @@ class AppServicesNav extends React.Component {
                         onClick={() => this.setActive("weather")}>
                         <WeatherButton temp={weatherData.currentTemperature} weatherIcon={weatherData.weatherIcon}/>
                         <div className={"t-small t-medium th-f-color tab-text even-spacing " + ((activeServiceTab == "weather") ? "" : "hide-tab")}>
-                            <div className="weather-item">
+                            <p className={weatherMissingData ? "waiting-message" : "hide-tab"}>
+                                {weatherMissingData ? waitingMessage : null}
+                            </p>
+                            <div className={weatherMissingData ? "hide-tab" : "weather-item"}>
```
jacobkeeler: `weatherData.high && weatherData.low` should probably be part of this condition, otherwise there is just a blank space before the rest of the data.

PR #171, opened 2019-06-17
Add logic for CloseApplication RPC

jacobkeeler:
Implements [#1931](github.com/smartdevicelink/sdl_core/issues/1931)

Also fixes issue where an unregistered application would bring the current app out of the foreground.


theresalech:
Resolves issue #175

PR #179, opened 2019-07-19
implement menu TILES

iCollin:
implements #165


src/js/InAppMenu.js
```diff
@@ -1,18 +1,24 @@
...
 
 export default class InAppMenu extends React.Component {
     constructor() {
         super();
     }
 
     render() {
+        const state = store.getState()
+        const app = state.ui[state.activeApp]
+        var layout = app ? (app.activeSubMenu ? app.menu.find(x => x.menuID === app.activeSubMenu).menuLayout : app.menuLayout) : "LIST";
```
JackLivio: Submenu souldnt have any impact on the top level app menu. If there is an active submenu, then the router would have the app in inAppList, not inAppMenu

src/js/containers/Menu.js
```diff
@@ -8,17 +9,12 @@ const mapStateToProps = (state) => {
     var menu = state.ui[activeApp].menu
     var theme = state.theme
     var data = menu.map((command) => {
-        var dataClass = null
-        if (command.cmdIcon) {
-            dataClass = 'with-icon'
```
JackLivio: Was the with-icon css class not being used?

src/js/containers/MenuIcon.js
```diff
@@ -24,6 +25,12 @@ const mapStateToProps = (state) => {
 const mapDispatchToProps = (dispatch) => {
     return {
         onSelection: (appID, path) => {
+            const state = store.getState()
+            const app = state.ui[state.activeApp]
+            if (app) {
+                app.activeSubMenu = null
```
JackLivio: There is an easier way to do this. Use the dispatch parameter.

```
dispatch(deactivateSubMenu(appID))
```

JackLivio: That should replace all changes here in this onSelection

JackLivio: Be sure to import the action

```
import { deactivateSubMenu } from '../actions'
```

src/js/reducers.js
```diff
@@ -506,6 +508,13 @@ function ui(state = {}, action) {
             var newState = { ...state }
             var app = newState[action.appID] ? newState[action.appID] : newAppState()
             return newState
+        case Actions.SET_GLOBAL_PROPERTIES:
+            var newState = { ...state }
+            var app = newState[action.appID] ? newState[action.appID] : newAppState()
+            if (action.menuLayout && action.menuLayout.length) {
+                app.menuLayout = action.menuLayout
+            }
```
JackLivio: I know this is not done everywhere, but i think this is better practice to include this line before returning newState.
```
newState[action.appID] = app
```

src/js/containers/Menu.js
iCollin: it was being used to decide between <svg> and , now this logic is done by the imageType

iCollin: github cut out my html from the comment. this class was used to differentiate between `<svg>` and `` but now imageType is used

src/js/reducers.js
iCollin: added in https://github.com/smartdevicelink/generic_hmi/pull/179/commits/a51b2fb95d7f7620d01a9566135817583345a4d1

src/js/containers/MenuIcon.js
iCollin: good call! changed in https://github.com/smartdevicelink/generic_hmi/pull/179/commits/a51b2fb95d7f7620d01a9566135817583345a4d1

src/js/InAppMenu.js
iCollin: That makes sens. Removed the submenu logic in https://github.com/smartdevicelink/generic_hmi/pull/179/commits/a51b2fb95d7f7620d01a9566135817583345a4d1

src/js/HScrollMenuItem.js
```diff
@@ -1,26 +1,32 @@
...
+            <Link
+                to={this.props.menuItem.link}
                 onClick={() => this.props.onSelection(this.props.appID, this.props.cmdID, this.props.menuID)}>
-                <div className="hscrollmenu-item__image">
-                    { graphic }
-                </div>
-                <div className="hscrollmenu-item__name">
-                    <p className="t-small t-light th-f-color">{menuItem.name}</p>
+                <div
+                    className={`hscrollmenu-item th-b-color th-tile-background-color`}>
```
ShobhitAd: Since the className is not being modified dynamically could simply change it to 
```suggestion
                    className="hscrollmenu-item th-b-color th-tile-background-color">
```

PR #190, opened 2019-08-08
Add app device name display in app list

ShobhitAd:
Adds support for issue [#2763](https://github.com/smartdevicelink/sdl_core/issues/2763)

Allows the user to differentiate between multiple apps with the same app Name using the displayed device name


iCollin:
Once https://github.com/smartdevicelink/generic_hmi/pull/179 is merged, developers will be able to change the layout of the apps menu to a `LIST`. We may want to add the device name display to that view as well.

PR #202, opened 2019-10-08
Base Widget Support

JackLivio:
This PR adds changes required to RPCs that were introduced by the widgets proposal.

This pr does not add widget functionality to the generic HMI, these are only minimal changes required for the HMI to work with Core 6.0.0.

Updates:
- Deprecate SetDisplayLayout
- Update show request to handle template configurations
- Add text supported to soft button capabilities
- Update Capabilities structure
- Add OnSystemCapabilityUpdated notification to core


iCollin:
I have a question about the `getWindowCapability(template)` function:

It's main client is BCController.js when handling the `OnAppRegistered` RPC, where it seems to be called with either `MEDIA` or `NON-MEDIA` template based on `rpc.params.application.isMediaApplication`.

Neither of these templates in DisplayCapabilities.js have `softButtonCapabilities` or `buttonCapabilities` within the `displayCapabilities` object.

Therefore, won't the return of this function always have a `null` `buttonCapabilities` and `softButtonCapabilities`?

JackLivio:
@iCollin Good catch. My mistake here was thinking softbutton capabilities was included in the displayCapabilities json object.

PR #210, opened 2019-10-15
make softbuttons in TILES_ONLY look more like HScrollMenu

iCollin:
This PR changes the look of SoftButtons in the TILES_ONLY template.

Updates:

- Images are now shown above the text instead of in-line with the text.
- Images can now be much larger to fill the tile button.
- If either the text or the image is not present, the containing HTML elements won't be rendered.


src/css/components/_tiles-only.scss
```diff
@@ -32,6 +32,14 @@
...
     position: relative;
+    @include flex-direction(column-reverse);
+    justify-content: space-evenly;
+}
+
+.tiles-only-template .soft-button-image {
+    height: auto;
+    max-width: 80%;
+    max-height: 60%;
 }
```
jacobkeeler:
```suggestion

.tiles-only-template .soft-button p {
    padding: 0;
}
```

This can be added to make it look better. Also, it appears that the text overflows outside of the box, so that might need to be looked at as well

iCollin: Added `padding: 0;` and `max-width: 80%;` in https://github.com/smartdevicelink/generic_hmi/pull/210/commits/8435f914b4ace356d458ca0bda593b2461c0ef6a so that text won't overflow.

PR #240, opened 2020-02-21
web engine support

### LINKED ISSUE #231: [SDL 0240] WebEngine support for SDL JavaScript
Proposal: [WebEngine support for SDL JavaScript](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0240-sdl-js-pwa.md)

> This proposal is adding a new transport to the SDL JavaScript library to support (progressive) web apps to run on a WebEngine or a browser.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/767

Steering Committee Decision:

>The Steering Committee voted to accept this proposal with the following revisions:
- When formatting `locales` in the `manifest.json` file, use standard [(RFC 5646)](https://tools.ietf.org/html/rfc5646), and convert locale strings when necessary.
- Do not specify the location of the manifest's <script> tag placement within the proposal.
- All revisions included in this PR: https://github.com/smartdevicelink/sdl_evolution/pull/888/files
> The proposal .md file was updated to reflect these revisions on 12/19/19.

iCollin:
Fixes #231 

This PR is **ready** for review.

Adds support for web engine apps and an App Store to the generic HMI.

Updates:
- Add `SetAppProperties` RPC
- Add `appStore` views, reducer and actions
- Add static file server (one for each web engine app running)

Tasks remaining:

- [x] remove extraneous logs
- [x] properly acknowledge `SetAppProperties` response
- [x] fix bug with uninstall and reinstall same app
- [x] fix bug with webengine app unregistering itself but still being in applist
- [x] add app store URL to flags
- [x] adjust snake case to camel case
- [x] cleanup appStore css
- [ ] look into api tell hmi when app is fully installed (optional) (HMI can poll manifest.js if necessary)
- [x] remove p tag for running web engine apps
- [x] test that PTU with vehicle modem still works (onReceivedPolicyUpdate thiscall?)
- [x] test css with different size names/descriptions/images
- [x] add an easy way to provide backendUrl to SimpleRPCClient
- [x] diagnose fileSystemController WS reconnect loop
- [x] find cause of onAppDeactivated(null, GENERAL)
- [x] installing both apps, running one removes one app from the app list(https://github.com/smartdevicelink/sdl_core/pull/3252)


src/js/AppHeader.js
```diff
@@ -35,7 +60,11 @@ class AppHeader extends React.Component {
         const themeClass = this.props.theme ? 'dark-theme' : 'light-theme';
         var modalClass = themeClass + " alertOverlay"
         var isShowingMenu = this.props.router.isActive('/inappmenu')
-        const icon = this.props.appIcon == 'false' ? (<div />) : <MenuIcon isShowingMenu={isShowingMenu}/> ;
+
+        const icon = this.props.appIcon == 'false' ? (<div />) : 
+            (this.props.appIcon == 'store' ? 
+                (this.props.router.isActive('/appstore') ? <AppStoreMenuIcon /> : <AppStoreIcon />)
```
JackLivio: I think 3 ternary operators in a row might be too many lol. Please break this up a bit.

src/js/AppHeader.js
```diff
@@ -35,7 +60,11 @@ class AppHeader extends React.Component {
         const themeClass = this.props.theme ? 'dark-theme' : 'light-theme';
         var modalClass = themeClass + " alertOverlay"
         var isShowingMenu = this.props.router.isActive('/inappmenu')
-        const icon = this.props.appIcon == 'false' ? (<div />) : <MenuIcon isShowingMenu={isShowingMenu}/> ;
+
+        const icon = this.props.appIcon == 'false' ? (<div />) :
```
JackLivio: Can appIcon be renamed to just `icon`

src/js/AppStore.js
```diff
@@ -0,0 +1,184 @@
...
+                let jsonEnd = manifestJS.lastIndexOf('}') + 1;
+                let manifest = JSON.parse(manifestJS.substring(jsonStart, jsonEnd))
+
+                var appProperties = {
+                    policyAppID: manifest.appId,
+                    enabled: true
+                }
+
+                let state = store.getState();
+                var appDirEntry = state.appStore.availableApps.find(x => x.policyAppID == manifest.appId);
```
JackLivio: Check that appDirEntry was found

src/js/AppHeader.js
iCollin: broken up in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/da5e2ba127e3604518601fd85ca7e8e7b210e3b0

src/js/AppHeader.js
iCollin: yes, done in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/da5e2ba127e3604518601fd85ca7e8e7b210e3b0

src/js/AppStore.js
```diff
@@ -0,0 +1,184 @@
...
+            if (!success || !params.appUrl) {
+                console.error('error encountered when installing app');
+            }
+
+            fetch(params.appUrl + 'manifest.js')
+                .then(x => x.text())
+                .then((manifestJS) => {
+                let jsonStart = manifestJS.indexOf('{');
+                let jsonEnd = manifestJS.lastIndexOf('}') + 1;
+                let manifest = JSON.parse(manifestJS.substring(jsonStart, jsonEnd))
```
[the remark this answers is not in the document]
iCollin: other junk like `var sdl_manifest = {}` and then `export default sdl_manifest;`

check added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/da5e2ba127e3604518601fd85ca7e8e7b210e3b0

src/js/AppStore.js
```diff
@@ -0,0 +1,196 @@
...
+        });
+    }
+
+    confirm() {
+        fileSystemController.subscribeToEvent('InstallApp', (success, params) => {
+            if (!success || !params.appUrl) {
+                console.error('error encountered when installing app');
+            }
+
+            fetch(params.appUrl + 'manifest.js')
```
JackLivio: Might want to add a "/" before manifest in case the appURL does not include that as the last character.

src/js/AppStore.js
```diff
@@ -0,0 +1,196 @@
...
+
+                let state = store.getState();
+                var appDirEntry = state.appStore.availableApps.find(x => x.policyAppID == manifest.appId);
+
+                if (!appDirEntry) {
+                    console.error('no app found in app store directory with policyAppID ', manifest.appId);
+                    return;
+                }
+
+                store.dispatch(appStoreAddAppPendingSetAppProperties(Object.assign(appDirEntry, {
```
JackLivio: I would remove the appStore prefixes from all of the app store related actions. The name is so long

src/js/AppStore.js
```diff
@@ -0,0 +1,196 @@
...
+        var pendingInstallApp = this.state.confirmApp ? this.state.confirmApp : {
+            name: null,
+            description: null,
+            iconUrl: null
+        };
+
+        return (
+            <div>
+                <Modal
+                isOpen={!!this.state.confirmID}
```
JackLivio: Bang bang?

src/js/AppStore.js
iCollin: check added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/da5e2ba127e3604518601fd85ca7e8e7b210e3b0

src/js/AppStore.js
iCollin: We decided it is mandatory for the `appUrl` to end in a `/`

src/js/AppStore.js
iCollin: to convert to boolean (react Modal will throw errors in console if it isn't a bool unfortunately)

if you like I could change it to:
```
this.state.confirmID ? true : false
```

src/js/AppStore.js
iCollin: I renamed this action in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/0f726e440962d70e67026eaa39e726cd0e342bf4, but I found the other action names to be reasonable. Let me know if you think appStore in any of the other names is redundant.

src/js/AppStore.js
JackLivio: Is that documented somewhere?

src/js/AppStore.js
JackLivio: No that is fine, i just didnt know you could do that

src/js/AppStore.js
```diff
@@ -0,0 +1,196 @@
...
+            });
+        });
+    }
+
+    return props;
+}
+
+export default connect(
+    mapStateToProps
+)(AppStore)
```
JackLivio: add line break at end

src/js/AppStoreMenu.js
```diff
@@ -0,0 +1,146 @@
...
+import { connect } from 'react-redux';
+import store from './store';
+import { addAppPendingSetAppProperties } from './actions'
+
+import AppHeader from './AppHeader';
+import StaticIcon from './Templates/Shared/StaticIcon';
+import ConfirmAlert from './ConfirmAlert';
+
+import BCController from './Controllers/BCController';
+import fileSystemController from './Controllers/FileSystemController';
```
JackLivio: Please change name to `FileSystemController`

src/js/AppStoreMenu.js
```diff
@@ -0,0 +1,146 @@
...
+                appName: name,
+                appIcon: icon,
+                appDescription: description
+            }
+        });
+    }
+
+    cancel() {
+        this.setState((state, props) => {
+            state.confirmID = null;
```
JackLivio: Should the other state params be set to null as well?

src/js/AppStoreMenu.js
```diff
@@ -0,0 +1,146 @@
...
+            return state;
+        });
+    }
+
+    confirm() {
```
JackLivio: Can this be named to confirmUninstall()

src/js/AppStoreMenu.js
```diff
@@ -0,0 +1,146 @@
...
+                contentLabel="Example Modal">
+                    <ConfirmAlert appID={this.state.confirmID} name={`Uninstall ${this.state.appName}?`}
+                        description={this.state.appDescription} iconUrl={this.state.appIcon}
+                        leftText="Cancel" leftCallback={this.cancel} rightText="Uninstall" rightCallback={this.confirm} />
+                </Modal>
+                <AppHeader icon='false' backLink="/appstore" menuName="STORE"/>
+                <div className="appstore-menu">
+                    {this.props.apps ? this.props.apps.map((app) => {
+
+                        var postFixIndex = 0;
```
JackLivio: Add a comment that this code is for calculating the app size that should be displayed

src/js/AppStoreMenu.js
```diff
@@ -0,0 +1,146 @@
...
+const mapStateToProps = (state) => {
+    return {
+        theme: state.theme,
+        apps: state.appStore.installedApps
+    }
+}
+
+export default connect(
+    mapStateToProps
+)(AppStoreMenu)
```
JackLivio: New line

src/js/Controllers/BCController.js
```diff
@@ -1,9 +1,10 @@
 import RpcFactory from './RpcFactory'
 import store from '../store'
-import { updateAppList, activateApp, deactivateApp, registerApplication, unregisterApplication, policyUpdate, onPutFile,  updateColorScheme, setAppIsConnected, onSystemCapabilityUpdated } from '../actions'
+import { updateAppList, activateApp, deactivateApp, registerApplication, unregisterApplication, policyUpdate, onPutFile,  updateColorScheme, setAppIsConnected, onSystemCapabilityUpdated, appStoreAppInstalled, appStoreAppUninstalled } from '../actions'
 import sdlController from './SDLController'
 import externalPolicies from './ExternalPoliciesController'
 import {flags} from '../Flags'
+import fileSystemController from './FileSystemController';
```
JackLivio: `FileSystemController`

src/js/WebEngineAppContainer.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export default class WebEngineAppContainer extends React.Component {
+    render() {
+        return (<div className="webengine-app-container">
+            <iframe src={this.props.iframeUrl} style={{ display: 'none' }} />
+        </div>);
+    }
+}
```
JackLivio: new line

src/js/actions.js
```diff
@@ -31,7 +31,13 @@ export const Actions = {
     ON_SYSTEM_CAPABILITY_UPDATED: "ON_SYSTEM_CAPABILITY_UPDATED",
     ON_APP_SERVICE_DATA: "ON_APP_SERVICE_DATA",
     ON_PUT_FILE: "ON_PUT_FILE",
-    SET_GLOBAL_PROPERTIES: "SET_GLOBAL_PROPERTIES"
+    SET_GLOBAL_PROPERTIES: "SET_GLOBAL_PROPERTIES",
+    UPDATE_AVAILABLE_APPSTORE_APPS: "UPDATE_AVAILABLE_APPSTORE_APPS",
+    UPDATE_INSTALLED_APPSTORE_APPS: "UPDATE_INSTALLED_APPSTORE_APPS",
+    APPSTORE_ADD_APP_PENDING_SET_APP_PROPERTIES: "APPSTORE_ADD_APP_PENDING_SET_APP_PROPERTIES",
```
JackLivio: Remove appstore prefixes here as well

src/js/containers/AppList.js
```diff
@@ -23,16 +25,33 @@ const mapStateToProps = (state) => {
...
+
+            if (!webEngineApp) {
+                sdlController.onAppActivated(appID)
+                return;
+            }
+
+            dispatch(webEngineAppLaunch(app.policyAppID, appID));
+
+            var activateAppOnceRegistered = setInterval(() => {
+                if (!app.isRegistered) {
```
JackLivio: Does this value change once it is registered? IE if it fails the first time can it succeed after?

src/js/actions.js
iCollin: removed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/WebEngineAppContainer.js
iCollin: added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/Controllers/BCController.js
iCollin: fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStoreMenu.js
iCollin: added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStoreMenu.js
iCollin: added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStoreMenu.js
iCollin: yes, also changed the cancel and the matching functions for install in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStoreMenu.js
iCollin: can't hurt, added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStoreMenu.js
iCollin: fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStore.js
iCollin: added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/AppStore.js
```diff
@@ -0,0 +1,196 @@
...
+                isOpen={!!this.state.confirmID}
+                className="alertModal app-body"
+                overlayClassName={modalClass}
+                contentLabel="Example Modal">
+                    <ConfirmAlert appID={this.state.confirmID} name={pendingInstallApp.name}
+                        description={pendingInstallApp.description} iconUrl={pendingInstallApp.iconUrl}
+                        leftText="Cancel" leftCallback={this.cancel} rightText="Download" rightCallback={this.confirm} />
+                </Modal>
+                <AppHeader icon='store' backLink="/" menuName="APPS"/>
+                {this.props.apps ? <HScrollMenu data={this.props.apps.map((app) => {
```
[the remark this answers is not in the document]
iCollin: fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/b79ab73fcfd760a30437c25d53d1c6665ee29526

src/js/containers/AppList.js
iCollin: Yes, with setInterval this function will be called every so often until the app is registered in which case it will stop waiting and send onAppActivated.

src/js/AppStore.js
```diff
@@ -0,0 +1,194 @@
...
+                isOpen={!!this.state.confirmID}
+                className="alertModal app-body"
+                overlayClassName={modalClass}
+                contentLabel="Example Modal">
+                    <ConfirmAlert appID={this.state.confirmID} name={pendingInstallApp.name}
+                        description={pendingInstallApp.description} iconUrl={pendingInstallApp.iconUrl}
+                        leftText="Cancel" leftCallback={() => this.cancelInstall()} rightText="Download" rightCallback={() => this.confirmInstall()} />
+                </Modal>
+                <AppHeader icon='store' backLink="/" menuName="APPS"/>
+                {<HScrollMenu data={this.props.apps.map((app) => {
```
JackLivio: Please move the map function outside of the return line.

src/js/AppStoreMenu.js
JackLivio: Also could you move the data size calculation outside of the return line. As a general rule I think for style reasons any logic or calculation that is slightly complicated should be calculated before the return.

JackLivio: Because the calculation is for each menu item, maybe the calculation should be moved to the AppStoreMenuItem render() method

src/js/ConfirmAlert.js
```diff
@@ -0,0 +1,33 @@
...
+                    </div>
+                    <div className="alert-button-2 th-f-color t-small t-light th-bg-color th-soft-buttons" 
+                        key={1} onClick={this.props.rightCallback}>
+                            {this.props.rightText}
+                    </div>
+                </div>
+            </div>
+        );
+    }
+}
```
JackLivio: new line please

src/js/Controllers/RpcFactory.js
```diff
@@ -1,6 +1,16 @@
 import {capabilities, getDisplayCapability} from './DisplayCapabilities.js'
 var rpcFactory_msgId = 5012
 class RpcFactory {
+    static BCSetAppProperties(properties) {
```
JackLivio: Is the BC prefix necessary?

src/js/Controllers/SDLController.js
```diff
@@ -3,7 +3,7 @@ import store from '../store'
 import { activateApp, setURLS, setPTUWithModem } from '../actions'
 import bcController from './BCController'
 import externalPolicies from './ExternalPoliciesController'
-import vehicleModem from './VehicleModemController'
+import fileSystemController from './FileSystemController'
```
JackLivio: FileSystemController

src/js/entry.js
```diff
@@ -68,11 +73,36 @@ class HMIApp extends React.Component {
...
+                        return (<WebEngineAppContainer key={app.policyAppID} policyAppID={app.policyAppID} iframeUrl={app.baseUrl + query} />);
+                    })
+                }
             </div>
         )
     }
     componentDidMount() {
         this.sdl.connectToSDL()
+
+        fileSystemController.connect(flags.FileSystemApiUrl).then(() => {
```
JackLivio: please change to FileSystemController

JackLivio: Find all and replace please

src/js/reducers.js
```diff
@@ -561,12 +581,74 @@ function system(state = {}, action) {
...
+            newState.installedApps = state.installedApps ? state.installedApps.map((app) => {
+                var appDirEntry = action.availableApps.find(x => x.policyAppID === app.policyAppID);
+                return appDirEntry ? Object.assign(appDirEntry, app) : app;
+            }) : [];
+            return newState;
+        case Actions.UPDATE_INSTALLED_APPSTORE_APPS:
+            var newState = { ...state };
+            newState.installedApps = action.installedApps.map((app) => {
+                var appDirEntry = newState.availableApps ? newState.availableApps.find(x => x.policyAppID === app.policyAppID) : {};
+                appDirEntry.runningAppId = 0;
```
JackLivio: Why assign to 0?

src/js/reducers.js
```diff
@@ -561,12 +581,74 @@ function system(state = {}, action) {
...
+                return Object.assign(appDirEntry, app);
+            });
+            return newState;
+        case Actions.ADD_APP_PENDING_SET_APP_PROPERTIES:
+            var newState = { ...state };
+            if(!newState.appsPendingSetAppProperties){
```
JackLivio: Please add spaces

JackLivio:
```suggestion
            if (!newState.appsPendingSetAppProperties) {
```

JackLivio: Example suggestion below

JackLivio:
General nice to haves if time permits:

- Installing an app should not make the app disappear from the app store list, instead it should be greyed out to indicate it was installed.
- If the above is implemented, then it would be nice to have a status change indicate when an app is downloading/installed.
- There is no app icon shown in the app list before activation after being downloaded. Is it possible to show the app icon from the app store if available?
- I would decrease the size of the app store icon to better match the line height of the "Apps" text.  30px width should do the trick.

src/js/entry.js
iCollin: fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/9420e8cd33100df3cba2420ebea12091814adcac

src/js/Controllers/SDLController.js
iCollin: fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/9420e8cd33100df3cba2420ebea12091814adcac

src/js/Controllers/RpcFactory.js
iCollin: removed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/9420e8cd33100df3cba2420ebea12091814adcac

src/js/ConfirmAlert.js
iCollin: added in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/9420e8cd33100df3cba2420ebea12091814adcac

src/js/AppStore.js
iCollin: done in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/9420e8cd33100df3cba2420ebea12091814adcac

src/js/reducers.js
ShobhitAd: Fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/4c6c9dbd8def6a6b0124c61c001b71e5e2dd6fc4

ShobhitAd: Fixed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/4c6c9dbd8def6a6b0124c61c001b71e5e2dd6fc4

src/js/reducers.js
iCollin: removed in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/0e2ac6d22aadfd72af1f339cf060802658c78e92

iCollin:
> * Installing an app should not make the app disappear from the app store list, instead it should be greyed out to indicate it was installed.
> * If the above is implemented, then it would be nice to have a status change indicate when an app is downloading/installed.

Why would you want an installed app to show up in the app store still?

Greyed out isn't currently supported in the generic HMI and CSS attributes cannot be passed down through the `HScrollMenu`. We would have to duplicate this component and all of its children components and then add greyed out logic to it which would be quite a big lift at this point.

It may be possible to add some status without greyed out by adding a `Downloading...` tag or similar while the app is pending install. I could look into this if we want it without greyed out.

> * There is no app icon shown in the app list before activation after being downloaded. Is it possible to show the app icon from the app store if available?

This is a side effect of not running the app until a user clicks on the app. It would be possible (and not too difficult) to pass an image struct such as this: ```{ imageType: 'url', image: 'site.com/img.png' }```
and implement logic in `HScrollMenuItem.js` to handle this new image type, but that means we are diverging from the API spec where we previously conformed.

If we decide to create a new `HScrollMenu` for adding greyed out elements we could add this without using a non-standard `imageType`.

> * I would decrease the size of the app store icon in the top right to better match the line height of the "Apps" text.  30px width should do the trick.

Good idea, shrunk in https://github.com/smartdevicelink/generic_hmi/pull/240/commits/c70bd89beb5d144f67aac7ff9070238c1ec78f71

JackLivio:
> Why would you want an installed app to show up in the app store still?

IMO I think the app disappearing is confusing and signals that something might have gone wrong.

> Greyed out isn't currently supported in the generic HMI and CSS attributes cannot be passed down through the HScrollMenu. We would have to duplicate this component and all of its children components and then add greyed out logic to it which would be quite a big lift at this point.

After reviewing the HMI API there is a param in the HMIApplication struct "greyOut". Looks like the applist needs to support a grey out parameter anyways. HScrollMenuItem could easily be extended to have a new property "greyOut".

> and implement logic in HScrollMenuItem.js to handle this new image type, but that means we are diverging from the API spec where we previously conformed.

In AppList.js mapStateToProps if there is no app.icon could you check if there is an app.iconUrl present? The AppStore HscrollMenu has no issue showing the app store icon via `image: app.iconUrl,`

> If we decide to create a new HScrollMenu for adding greyed out elements we could add this without using a non-standard imageType.

Should not be necessary to create a new HScrollMenu implementation since the regular applist should support greyOut.

iCollin:
> IMO I think the app disappearing is confusing and signals that something might have gone wrong.

Now it renames the app to `Installing...` while the app is being installed and then greys out the app in the store. Let me know what you think of this implementation.

> In AppList.js mapStateToProps if there is no app.icon could you check if there is an app.iconUrl present? The AppStore HscrollMenu has no issue showing the app store icon via `image: app.iconUrl,`

I now added a check, if there is no icon, mapStateToProps will look through the appStore.installedApps for a matching policyAppId and pass on the iconUrl of any matching app there. For some reason it seems that `img src=URL` is rendering differently than `img src=FILE`.

I tried a couple things, but the difference is minimal. I will keep looking for a solution to the rendering differences.

PR #241, opened 2020-02-28
Fix/Web engine backend connection errors

iCollin:
Issues
- if one app is still installing and you try to install another, there is a javascript error where both icons combine( Fixed in https://github.com/smartdevicelink/generic_hmi/pull/241/commits/9b66773ddb39c8f3ba4487fa7e7cc1b80cf77e9f)
- do not show app store unless local server is running ( Fixed in https://github.com/smartdevicelink/generic_hmi/pull/241/commits/b7bfb5eec2d320414a22ba3a351402670479f6ec)
- null check if server isn't running


src/js/AppStore.js
```diff
@@ -149,7 +149,7 @@ class AppStore extends React.Component {
                     image: app.iconUrl,
                     appID: app.policyAppID,
                     cmdID: app.package_url,
-                    name: app.pendingInstall ? 'Installing...' : app.name,
+                    name: app.pendingInstall ? `Installing ${app.name}...` : app.name,
```
iCollin: Why change this? Installing + the app name will very rarely fit on the app tile and then the user will be left wondering what the heck it says haha. I chose Installing... because I figured they would remember which app they clicked install on or could tell by the icon and the app tile could still look clean.

src/js/Controllers/BCController.js
```diff
@@ -86,7 +86,13 @@ class BCController {
         switch (methodName) {
             case "SetAppProperties":
                 let success = (rpc.result.resultCode == 'SUCCESS')
-                let entry = store.getState().appStore.appsPendingSetAppProperties[0]
+                let appsPendingSetAppProperties = store.getState().appStore.appsPendingSetAppProperties;
+                if(!appsPendingSetAppProperties || appsPendingSetAppProperties.length == 0){
+                    console.error("SetAppProperties Response: no apps in pending queue");
```
iCollin: was this case hit? this shouldn't be possible.

src/js/AppHeader.js
```diff
@@ -64,7 +64,11 @@ class AppHeader extends React.Component {
         var icon = this.props.icon == 'false' ? (<div />) : (<MenuIcon isShowingMenu={isShowingMenu}/>);
 
         if (this.props.icon == 'store') {
-            icon = this.props.router.isActive('/appstore') ? (<AppStoreMenuIcon />) : (<AppStoreIcon />);
+            if(this.props.isAppStoreConnected){
```
iCollin:
```suggestion
            if (this.props.isAppStoreConnected) {
```
you dont like spaces? i think it would be nice to keep the style consistent

src/js/AppStore.js
ShobhitAd: Just wanted the tiles to have different names that's all. Without this change, if you try to install multiple apps simultaneously(meaning multiple tile have the same value `Installing...`) it cause an error 
```js
Warning: flattenChildren(...): Encountered two children with the same key, `undefinedInstalling...`. Child keys must be unique; when two children share a key, only the first child will be used.
    in div (created by HScrollMenu)
    in HScrollMenu (created by AppStore)
    in div (created by AppStore)
    in AppStore (created by Connect(AppStore))
    in Connect(AppStore) (created by RouterContext)
    in RouterContext (created by Router)
    in Router
    in div (created by HMIApp)
    in div (created by HMIApp)
    in div (created by HMIApp)
    in HMIApp (created by Connect(HMIApp))
    in Connect(HMIApp)
...
```
It doesn't have to be the app name any unique identifier would be fine, the appName was just the most appropriate info to display to the user and the easiest to get.

src/js/Controllers/BCController.js
ShobhitAd: Just a null check. I wasn't able to hit this case when I tested but there might be a way to trigger it

src/js/AppHeader.js
ShobhitAd: lol I wouldn't go that far, just not in the habit of adding them consistently. This one and a few others fixed in https://github.com/smartdevicelink/generic_hmi/pull/241/commits/a6eedda259a03dff1fed10ce568e10a4aa497436

src/js/AppStore.js
iCollin: `Without this change,` fixing the error is just a side-effect of your change.

the key of the div is what matters, not the name shown in the UI: ```key={menuItem.cmdID + menuItem.name}```

ShobhitAd: Oh I think I understand what you mean about the key. Fixed in https://github.com/smartdevicelink/generic_hmi/pull/241/commits/c16b4174d78769d08df67c10ae784653fbb23407

src/js/actions.js
```diff
@@ -326,17 +327,24 @@ export const setGlobalProperties = (appID, menuLayout) => {
...
+
 export const updateAvailableAppStoreApps = (availableApps) => {
     return {
         type: Actions.UPDATE_AVAILABLE_APPSTORE_APPS,
         availableApps: availableApps
     };
 }
 
-export const updateInstalledAppStoreApps = (installedApps) => {
+export const updateInstalledAppStoreApps = (installedApp) => {
```
iCollin: `updateInstalledAppStoreApps` and `UPDATE_INSTALLED_APPSTORE_APPS` should be changed too if we're going to change the parameter of the action.

ShobhitAd: @iCollin I think the name `UPDATE_INSTALLED_APPSTORE_APPS` should still apply since it's updating the **list of installed app store apps** by either adding a new `installedApp` or modifying an existing one.

PR #249, opened 2020-03-31
Add long button press support

### LINKED ISSUE #143: Add ability to detect long presses on subscribe buttons
Currently the generic hmi only sends a `ButtonPressMode` of `Short` when a subscribe button is selected via the UI. The generic hmi should be able to distinguish between long and short presses and send the appropriate `ButtonPressMode`.

JackLivio:
Add support for long button presses. If the user holds the button for 3 seconds, then a long button press will be sent.

I intentionally left the onClick code commented out in case someone's hmi only supports the onclick dom handler.

Fixes #143


src/js/SoftButtons.js
```diff
@@ -14,31 +14,43 @@ export default class SoftButtons extends React.Component {
             items = softButtons.map((softButton, index) => {
                 return (<div className="soft-button-tile-large th-f-color t-small t-light th-bg-color th-soft-buttons" 
                             key={softButton.softButtonID}
-                            onClick={() => this.props.onButtonPress(this.props.appID, softButton.softButtonID, "CUSTOM_BUTTON")}>
+                            /*onClick={() => this.props.onButtonPress(this.props.appID, softButton.softButtonID, "CUSTOM_BUTTON")}*/
+                            onMouseDown={() => this.props.onButtonDown(this.props.appID, softButton.softButtonID, "CUSTOM_BUTTON")}
+                            onMouseUp={() => this.props.onButtonUp(this.props.appID, softButton.softButtonID, "CUSTOM_BUTTON")}                           
+                        >
```
jacobkeeler: Is it possible to return this in a function? It seems like this code is repeated in quite a few places

jacobkeeler: Also, remove commented `onClick` function

src/js/Controllers/UIController.js
```diff
@@ -291,6 +291,18 @@ class UIController {
         button.mode = "SHORT"
         this.listener.send(RpcFactory.OnButtonPressNotification(appID, button))
     }
+    onLongButtonPress(appID, buttonID, buttonName) {
+        var button = {
+            name: buttonName,
+            mode: "BUTTONDOWN",
+            customButtonID: buttonID
+        }
```
jacobkeeler: ButtonEvent messages should be sent in their respective functions, not all at once

src/js/Controllers/DisplayCapabilities.js
```diff
@@ -1,7 +1,7 @@
 let softButtonCapability = {
 	"shortPressAvailable": true,
-	"longPressAvailable": false,
-	"upDownAvailable": false,
+	"longPressAvailable": true,
+	"upDownAvailable": true,
```
jacobkeeler, line 4: Will need to modify the values of the button capabilities in the MEDIA template as well

JackLivio: Updated

NicoleYarroch:
I tested the long and short press on both a subscribe button and soft button and it worked as expected.

PR #250, opened 2020-03-31
Make album art consistent size

### LINKED ISSUE #131: CSS issue when isTemplate is set to true
I noticed some css issues when setting `isTemplate` to `true` for an image. The image occupies more space than its actual size:

This is how the image looks when isTemaplate is set to `true`. Notice that it takes up a big square area even though the image is rectangular.

screen shot 2018-10-29 at 3 17 34 pm

This is how it looks when isTemaplte is set to `false`

screen shot 2018-10-29 at 3 18 08 pm

JackLivio:
Add extra div to allow the album artwork div take up the same size no matter if it is a template image or not.

Fixes #131


iCollin:
This does what you say it will do and code looks good.

But isn't the bug that "he image occupies more space than its actual size"?

JackLivio:
Decided that the area taken up from an album art image should always be a square (because album artworks are square). Also template images are redrawn onto a canvas that are set to the size of the parent div so it is difficult to resize the canvas to the images natural size.

PR #253, opened 2020-04-02
move vscrollmenuitem icon to the left side

### LINKED ISSUE #233: The menu icon's alignment is off when the menu has multiple lines
# Bug Report
When a menu's title is long enough to fit on multiple lines, the position of the icon is a little off. 
Menu Icon Alignment

## Reproduce
1. Create a PICS that has at least one choice set item. It should have an icon and the title should be long enough to span two lines.  

### Expected Result
There should be padding between the icon and the right arrow. The icon could be vertically aligned in the row as well. 

### Actual Result
The icon touches the right arrow.

iCollin:
Fixes #233 

Changelog:
Align vscrollmenuitem icon vertically
Move vscrollmenuitem icon to the left side
Use flex to make moving these components simpler


src/css/components/_v-scroll-menu-item.scss
```diff
@@ -1,43 +1,67 @@
...
+        height: 100%;
+    }
+
     &__image {
-        @include size(150px);
         margin: 25px auto 12px;
 
-
+        flex: 1;
+        align-self: flex-start;
```
ShobhitAd, line 26:
```suggestion
```
Removing this allows the image to be centered vertically

src/css/components/_v-scroll-menu-item.scss
```diff
@@ -1,43 +1,67 @@
...
         border-radius: 4px;
 
         img {
+            display: block;
+            margin:auto;
+
             max-width: 100%;
+            height: 45px;
```
ShobhitAd, line 34:
```suggestion
            max-height: 45px;
```
Probably shouldn't have a fixed height for all images

src/css/components/_v-scroll-menu-item.scss
```diff
@@ -1,43 +1,67 @@
...
+
+        div {
+            margin: auto;
+
+            svg {
+                display: block;
+                margin:auto;
+
+                max-width: 100%;
+                height: 45px;
```
ShobhitAd, line 45:
```suggestion
                max-height: 45px;
```

iCollin: Please see https://github.com/smartdevicelink/generic_hmi/pull/253/commits/f76c98926f6963a656f85be07aa1ee186e4f7b68

src/css/components/_v-scroll-menu-item.scss
iCollin: Change adopted in https://github.com/smartdevicelink/generic_hmi/pull/253/commits/f76c98926f6963a656f85be07aa1ee186e4f7b68

src/css/components/_v-scroll-menu-item.scss
iCollin: align-self is required with space-between to not have additional margin on the left. I have added changes to achieve the same effect in https://github.com/smartdevicelink/generic_hmi/pull/253/commits/f76c98926f6963a656f85be07aa1ee186e4f7b68

NicoleYarroch:
Our custom icons for the Livio Music player are getting a lot of extra left and right padding. 
icon padding in menu

NicoleYarroch:
I am still seeing the same exact issue with the extra padding. The images we are uploading are 128x128 pixels.

PR #254, opened 2020-04-02
 put icon above name of soft button tiles

### LINKED ISSUE #242: Only part of the text in a soft button that supports both text and images is shown.
# Bug Report
Only part of the text in a soft button that supports both text and images is shown.

1. Create an app with `appType` of `COMMUNICATION` and a main screen template of `GRAPHIC_WITH_TEXTBUTTONS`.
1. Create 4 soft buttons that support both text and images. 
1. Display the soft buttons using a `Show` request.

### Expected result
All the soft buttons show both the image and the text.

### Actual result
Only a sliver of the text is shown to the left of the button. In the screenshot, the top left soft button is supposed to show "Call" under the image. The highlighted box shows that you can only see part of the "C". You can see part of the text on the left side of each button. 

Generic_HMI_text_and_image_soft_button_bug_day_mode

### More Information
**Bug seen on:**  Manticore (SDL Core v6.0.1 and Generic HMI v0.7.2.) and the develop branch of Generic_HMI (commit 0ae8e3ea6d475651a83ff44da23f1d434b4266e7)
**Bug seen on web browsers:** Safari, Chrome, and Chromium

iCollin:
Fixes #242

Changelog:
update flex direction of soft button tile to column reverse
update DYNAMIC image container so that they will be the same size as static images


NicoleYarroch:
All image-only, text-only, and image-and-text buttons look good. 

I did notice that:
1. If you have a long line of text, the text extends outside the bounds of the soft button.
1. The soft buttons are not equal height if one row of soft buttons contains text-only buttons and another row has soft buttons with images. I think it would look better if the buttons were all of equal height. 

soft buttons

NicoleYarroch:
Looks good.

PR #263, opened 2020-04-27
dont move to an app's template until it is actually activated

iCollin:
Wait for a positive response from SDL core to leave the app menu.


src/js/HScrollMenu.js
```diff
@@ -1,19 +1,36 @@
...
+            return;
+        }
+
+        this.props.onSelection(appID, cmdID, menuID);
+        store.dispatch(setPendingAppLaunch(appID));
     }
 
     render() {
         let menuItems = this.props.data.map((menuItem) => {
+            if (menuItem.appID === this.props.pendingAppLaunch) {
```
jacobkeeler, line 23: Seems that this sort of logic should be in the `AppList` file, rather than the base menu class.

src/js/containers/AppList.js
```diff
@@ -36,13 +36,13 @@ const mapStateToProps = (state) => {
             name: name,
             devicename: devicename,
             image: icon,
-            link: '/' + link,
+            link: undefined,
```
jacobkeeler, line 39: What was the purpose of this change?

iCollin: This prevents the HMI from moving to a blank media template immediately when an app is clicked on.

src/js/containers/AppList.js
```diff
@@ -30,25 +30,39 @@ const mapStateToProps = (state) => {
...
             cmdID: app,
-            greyOut: app.greyOut
+            greyOut: greyOut
         }
     })
 
-    return {data: data}
+    return {data: data, pendingAppLaunch: state.pendingAppLaunch}
```
iCollin, line 52:
```suggestion
    return {data: data}
```

There is no reason for HScrollMenu to receive this prop.

PR #266, opened 2020-04-30
Fix PROPRIETARY update flow

jacobkeeler:
Don't send URLs in OnSystemRequest, cycle URLs in EXTERNAL_PROPRIETARY retry sequence


src/js/Controllers/ExternalPoliciesController.js
```diff
@@ -62,11 +62,10 @@ class ExternalPoliciesController {
         }
     }
     onPackMessage(evt) {
-        bcController.onSystemRequest(this.sysReqParams.fileName, this.sysReqParams.urls)
+        bcController.onSystemRequest(this.sysReqParams.fileName, this.sysReqParams.urls[0].url)
```
iCollin, line 65: will these properties always be there? do we need any nullcheck?

src/js/Controllers/ExternalPoliciesController.js
```diff
@@ -88,7 +87,8 @@ class ExternalPoliciesController {
             
             this.policyUpdateRetryTimer = setTimeout(
                 function() {
-                    bcController.onSystemRequest(this.sysReqParams.policyUpdateFile, this.sysReqParams.urls)
+                    var endpoint = this.sysReqParams.urls[retryCount % this.sysReqParams.urls.length];
```
iCollin, line 90: will there always be urls?

PR #272, opened 2020-05-21
Fix/Webengine server host config

ShobhitAd:
At the moment the webengine backend server only works correctly for localhost(`127.0.0.1`).

This PR adds changes for the host to be easily configured in the main function of the `start_server.py` script

To configure the host change this line
```
- backend_server = WSServer('localhost', 8081, RPCService)
+ backend_server = WSServer('<hostname>', 8081, RPCService)
```

### DIFF
--- a/tools/start_server.py
+++ b/tools/start_server.py
@@ -321,7 +321,7 @@ def main():
 
   host = str(sys.argv[1])
   port = int(sys.argv[2])
-  WebEngineManager.WEBENGINE_REMOTE_HOST = str(sys.argv[3]) if (len(sys.argv) > 3 and str(sys.argv[3]).strip() != "") else host
+  WebEngineManager.WEBENGINE_REMOTE_HOST = str(sys.argv[3]) if len(sys.argv) > 3 else host
   WebEngineManager.next_available_port = int(sys.argv[4]) if (len(sys.argv) > 4 and int(sys.argv[4]) != 0) else WebEngineManager.next_available_port
 
   backend_server = WSServer(host, port, RPCService)
--- a/tools/start_server.py
+++ b/tools/start_server.py
@@ -38,6 +38,7 @@ from zipfile import ZipFile
 import subprocess
 import uuid
 import time
+import sys
 
 class WSServer():
   def __init__(self, _host, _port, _service_class=None):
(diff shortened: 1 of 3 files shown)


tools/start_server.py
```diff
@@ -249,11 +250,11 @@ def start_app_file_server(self, _app_storage_folder):
     secret_key = str(uuid.uuid4())
     print('\033[2mPort is %s, secret key is %s\033[0m' % (port, secret_key))
 
-    process = subprocess.Popen(['python3', '../../tools/file_server.py', str(port), secret_key], cwd=_app_storage_folder)
+    process = subprocess.Popen(['python3', '../../tools/file_server.py', str(self.WEBENGINE_HOST), str(port), secret_key], cwd=_app_storage_folder)
     time.sleep(1)
 
     WebEngineManager.next_available_port += 1
-    return {'process': process, 'url': 'http://localhost:%s/%s/' % (port, secret_key)}
+    return {'process': process, 'url': 'http://%s:%s/%s/' % (self.WEBENGINE_HOST, port, secret_key)}
```
iCollin, line 257: `self.WEBENGINE_HOST` will be the client's IP, not the file server IP. This will only work on localhost.

I think this value should just be user defined at the top of the file or as an argument and default to 'localhost'

ShobhitAd: Added hostname config option in https://github.com/smartdevicelink/generic_hmi/pull/272/commits/6ef0b3ca6d19109c44247b0d88aab322f642f98c

tools/start_server.py
```diff
@@ -319,6 +321,9 @@ def main():
 
   host = str(sys.argv[1])
   port = int(sys.argv[2])
+  WebEngineManager.WEBENGINE_REMOTE_HOST = str(sys.argv[3]) if (len(sys.argv) > 3 and str(sys.argv[3]).strip() != "") else host
```
iCollin, line 324:
```suggestion
  WebEngineManager.WEBENGINE_REMOTE_HOST = str(sys.argv[3]) if len(sys.argv) > 3 else host
```

is this really necessary? Someone would have to put whitespace in quotes in order to hit this code

ShobhitAd: good catch. That check wouldn't work as expected

PR #282, opened 2020-06-18
Feature/additional sub menus

### LINKED ISSUE #279: [SDL 0148] Template Improvements: Additional SubMenus
Proposal: [Template Improvements: Additional SubMenus](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0148-template-additional-submenus.md)

> Currently, SDL is by far the most restrictive among its competitors for drilling down into menus. It only allows the top menu and one submenu (additional submenus can be "faked" using Choice Sets). CarPlay, for example, allows at least 3 levels of drilling, this was the most I could find in the apps I use (in addition to tabs at the top, which is a kind-of 4th level). CarPlay allows as much drilling as desired by the user in the car, but prevents scrolling while moving. For future updates, we should allow further menu depth. For driver distraction reasons, a later proposal will allow restriction of menu depth while driving to replicate current restrictions if desired.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/423

Steering Committee Decision:

> The Steering Committee has voted to accept this proposal with the following revisions: update proposal to only include details regarding "Allowing Submenu's with parentID's that refer another Submenu." A separate proposal will be entered to discuss "Allowing Core to define driver distraction parameters for HMI."

>When discussing submenus, a few additional points were made:

>Right now, this is only considered for submenus, but driver distraction guidelines could be extended to include choice sets for interactions as well (including list length)
Proposal outlines the same behavior as is currently in place for submenus, but extends to allow a submenu to have a submenu

>The proposal .md was updated on March 22, 2018 to include the agreed upon revisions.

JackLivio:
Adds functionailty for additional submenu depth and driver distraction limitations on command list length and submenu depth.

Fixes #279 #280


src/js/containers/SubMenu.js
```diff
@@ -38,33 +46,47 @@ const mapStateToProps = (state) => {
...
     var menu = app.menu
     var activeSubMenu = app.activeSubMenu
-    if (!activeSubMenu || !menu) {
-        return {data: [], isPerformingInteraction: false, theme: theme}
-    }
-    data = menu.find((test) => {
-        return test.menuID === activeSubMenu
-    }).subMenu.map((command) => {
+    var ddState = state.ddState;
+    var subMenuData = SubmenuDeepFind(menu, activeSubMenu, 0).subMenu.subMenu.map((command, index) => {
```
jacobkeeler, line 50: The HMI crashes on this line when I select anything in a TILED submenu or any command within a submenu

```
TypeError: Cannot read property 'subMenu' of null
    at Function.Mn [as mapToProps] (SubMenu.js:50)
    at r (wrapMapToProps.js:41)
    at Function.r.mapToProps (wrapMapToProps.js:50)
    at r (wrapMapToProps.js:41)
    at selectorFactory.js:22
    at connectAdvanced.js:319
    at Object.useMemo (react-dom.production.min.js:4005)
    at t.useMemo (react.production.min.js:411)
    at O (connectAdvanced.js:304)
    at qi (react-dom.production.min.js:3667)
```

src/js/containers/SubMenu.js
```diff
@@ -2,8 +2,10 @@ import { connect } from 'react-redux'
 import VScrollMenu from '../VScrollMenu'
```
jacobkeeler, line 2: Overall design comments:

* Seems this might be an existing issue, but there really is no indication of which menu items are submenus in a TILED layout
* As far as implementing driver distraction limitations, I think we should either grey out the options that go deeper than the limit or omit them from the list entirely, leaving only commands at the final allowed level. It would be nice if there were some indication that items were omitted or disabled for safety purposes.

JackLivio: Made updates to this issue in 
06c0b3d

src/js/containers/SubMenu.js
JackLivio: Fixed in 
dce1586

src/js/Controllers/DisplayCapabilities.js
```diff
@@ -489,7 +489,11 @@ let capabilities = {
 			},
 			"phoneCapability": {
 				"dialNumberEnabled": false
-			}
+			},
+			"driverDistractionCapability": {
+                "menuLength": 3,
```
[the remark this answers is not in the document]
JackLivio: Converted all to spaces to match formatting in the rest of the project.

PR #284, opened 2020-06-26
Feature/Supported Character Sets

ShobhitAd:
Implements #283

### DIFF
--- a/src/js/Controllers/DisplayCapabilities.js
+++ b/src/js/Controllers/DisplayCapabilities.js
@@ -32,10 +32,7 @@ let screenParams = {
 	}
 }
 
-function textField(name, width, rows, characterSet) {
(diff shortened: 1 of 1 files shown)


src/js/Controllers/DisplayCapabilities.js
```diff
@@ -32,12 +32,13 @@ let screenParams = {
 	}
 }
 
-function textField(name, width, rows) {
+function textField(name, width, rows, characterSet) {
 	width = width ? width : 500
 	rows = rows ? rows : 1
+	characterSet = characterSet ? characterSet : "UTF_8"
```
iCollin, line 38:
```suggestion
function textField(name, width=500, rows=1, characterSet="UTF_8") {
```

what do you think about default params?

PR #285, opened 2020-06-30
Dynamic Menu and Image Capabilities

### LINKED ISSUE #234: [SDL 0268] Main Menu Updating and Pagination
Proposal: [Main Menu Updating and Pagination](https://github.com/smartdevicelink/sdl_evolution/blob/master/proposals/0268-main-menu-updating.md)

> This proposal seeks to add the ability to update the main menu with artworks and sub-menus on demand so that we can improve the performance and usability of the menu.

Review: https://github.com/smartdevicelink/sdl_evolution/issues/883

JackLivio:
Fixes #234 

Based off of the additional sub menus pr.


src/js/Templates/Shared/Image.js
```diff
@@ -67,6 +70,12 @@ export default class Image extends React.Component {
     }
 
     onError(event) {
+        const state = store.getState();
+        const activeApp = state.activeApp;
+        if (activeApp) {
+            const fileName = path.basename(this.props.image);
+            UIController.onUpdateFile(activeApp, fileName);
+        }
```
jacobkeeler, line 78: Oddly I only get this notification for my `appIcon`, even if I send a nonexistant image in an `AddSubMenu` request

src/js/containers/Menu.js
```diff
@@ -22,6 +23,11 @@ const mapStateToProps = (state) => {
             if (ddState === true && menuDepth === 1) {
                 enabled = false;
             }
+
+            if (command.subMenu.length === 0) {
+                // Found and empty submenu, ask app to send add commands
+                UIController.onUpdateSubMenu(activeApp, command.menuID);
+            }
```
jacobkeeler, line 30: I don't see this trigger go off, `OnUpdateSubMenu` is not sent when an empty submenu is shown or opened

JackLivio: Should be triggered when one level higher than an empty sub menu. Idea is to have the items populated that are one level deeper than the current menu depth.

src/js/Templates/Shared/Image.js
JackLivio: I am not seeing that issue. Can you post the message you are sending?

Here is the android code i was using for an addSubMenu with no image:


```
                AddSubMenu as = new AddSubMenu();
                as.setMenuName("sub");
                as.setMenuID(70);
                Image i = new Image();
                i.setImageType(ImageType.DYNAMIC);
                i.setValue("fake_file.png");
                as.setMenuIcon(i);
                sdlManager.sendRPC(as);
```

I only receive the onerror for the missing submenu icon when viewing the menu: 
Screen Shot 2020-08-05 at 12 15 36 PM

Do you see an onerror in the console for the app icon?

src/js/containers/Menu.js
JackLivio: Using the same code above, here is the onsubMenu triggered after opening the main menu:

Screen Shot 2020-08-05 at 12 18 21 PM

jacobkeeler:
An additional issue I ran into, it looks like the HMI will crash if an app is disconnected while on the menu screen. This is the error I get in such a case:
```
scheduler.production.min.js:96 Uncaught TypeError: Cannot read property 'menu' of undefined
    at Function.Pn [as mapToProps] (Menu.js:13)
    at r (wrapMapToProps.js:41)
    at selectorFactory.js:44
    at m (selectorFactory.js:58)
    at selectorFactory.js:63
    at e.d (connectAdvanced.js:64)
    at e.t.handleChangeWrapper (Subscription.js:95)
    at Subscription.js:23
    at el (react-dom.production.min.js:5919)
    at Object.notify (Subscription.js:19)
```

src/js/containers/Menu.js
jacobkeeler: Does this still work for nested submenus on your end? I'm still seeing the issue

NicoleYarroch:
I noticed when testing that images that are not templated do not show up in the menu even though they have been uploaded to core and I can see them in the storage folder. Scrolling does not make the non-templated images show up. The only way I can get the images to show up is to close the menu and re-open it. If I use templated images, everything seems to work as expected.

PR #291, opened 2020-07-27
webengine projection

iCollin:
implement https://github.com/smartdevicelink/generic_hmi/issues/275

this PR is ready for review!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -7,6 +7,7 @@ import './css/main.scss';
 // import react and js
 import MediaPlayer from './js/MediaPlayer';
 import NonMedia from './js/Templates/NonMedia/NonMedia'
+import WebView from './js/Templates/WebView/WebView'
 import LargeGraphicOnly from './js/Templates/LargeGraphicOnly/LargeGraphicOnly'
 import LargeGraphicWithSoftbuttons from './js/Templates/LargeGraphicWithSoftbuttons/LargeGraphicWithSoftbuttons'
 import GraphicWithTextButtons from './js/Templates/GraphicWithTextButtons/GraphicWithTextButtons'
@@ -78,7 +79,10 @@ class HMIApp extends React.Component {
                 {
                     this.props.webEngineApps.map((app) => {
                         let query = `?sdl-host=${flags.CoreHost}&sdl-port=${flags.CoreWebEngineAppPort}&sdl-transport-role=${app.transportType.toLowerCase()}-server`;
-                        return (<WebEngineAppContainer key={app.policyAppID} policyAppID={app.policyAppID} iframeUrl={app.appUrl + app.entrypoint + query} />);
+                        return (<WebEngineAppContainer key={app.policyAppID} 
+                            style={app.style}
+                            policyAppID={app.policyAppID} 
+                            iframeUrl={app.appUrl + app.entrypoint + query} />);
                     })
                 }
             </div>
@@ -138,6 +142,7 @@ ReactDOM.render((
             <Route path="/" exact component={HMIMenu} />
             <Route path="/media" component={MediaPlayer} />
             <Route path="/nonmedia" component={NonMedia} />
+            <Route path="/webview" component={WebView} />
             <Route path="/large-graphic-only" component={LargeGraphicOnly} />
             <Route path="/large-graphic-with-softbuttons" component={LargeGraphicWithSoftbuttons} />
             <Route path="/graphic-with-text-buttons" component={GraphicWithTextButtons} />
--- a/src/index.js
+++ b/src/index.js
@@ -94,6 +94,18 @@ class HMIApp extends React.Component {
     }
     componentDidMount() {
         this.sdl.connectToSDL()
+        if (window.performance.memory) { // chromium --enable-precise-memory-info
+            this.memoryUsageInterval = setInterval(function(app) {
+                var mem = window.performance.memory;
+                if (mem.usedJSHeapSize / mem.jsHeapSizeLimit > 0.75) {
+                    for (var app of this.webEngineApps) {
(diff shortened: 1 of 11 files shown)


src/index.js
```diff
@@ -78,14 +79,33 @@ class HMIApp extends React.Component {
...
     }
     componentDidMount() {
         this.sdl.connectToSDL()
+        if (window.performance.memory) { // chromium --enable-precise-memory-info
+            this.memoryUsageInterval = setInterval(function(self) {
+                var mem = window.performance.memory;
+                if (mem.usedJSHeapSize / mem.jsHeapSizeLimit > 0.75) {
+                    for (var app of self.props.webEngineApps) {
+                        if (app.runningAppId && app.style.position === 'absolute') {
+                            bcController.onExitApplication('RESOURCE_CONSTRAINT', app.runningAppId);
```
JackLivio, line 103: If the app is taking up too much memory, should the HMI also "delete" the iframe? Does that happen indirectly after the OnExitApplication message is sent?

src/js/AppStore.js
```diff
@@ -55,11 +55,25 @@ class AppStore extends React.Component {
                     return;
                 }
 
+                var style = {};
+                if (manifest.category === 'WEB_VIEW') {
+                    style = {
+                        position: 'absolute',
+                        width: '960px',
```
JackLivio, line 62: The size of the web view needs to take into account the dynamic height and width parameters set in the config and used by the scss files. It might be cleaner to define these styles in an scss file and pass a className within the js code instead.

src/js/Controllers/UIController.js
```diff
@@ -295,6 +297,11 @@ class UIController {
         this.listener.send(RpcFactory.UIPerformInteractionResponse(choiceID, appID, msgID))
     }
     onSystemContext(context, appID) {
+        if (context === 'MAIN' || context === 'ALERT') {
+            store.dispatch(appStoreShowWebView());
```
JackLivio, line 301: Should you check that the current route is the appstore/webview before dispatching these actions?

src/js/Templates/WebView/WebView.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import AppHeader from '../../containers/Header';
+
+class WebView extends React.Component {
+    render() {
+        return (
+            <div className="webView">
+                <AppHeader backLink="/" menuName="Apps" icon="false"/>
```
JackLivio, line 8: I assune WebengineAppContainer was not included in this component because the iframe needs to be running when not in view?

src/js/reducers.js
```diff
@@ -434,6 +434,9 @@ function ui(state = {}, action) {
                 case "NON-MEDIA":
                     app.displayLayout = "nonmedia"
                     break
+                case "WEB_VIEW":
+                    app.displayLayout = "webview"
```
JackLivio, line 438: I think web-view would be more inline with the naming conventions, except nonmedia is kind of ruining my claim.

src/index.js
iCollin: This will indirectly remove the iframe after OnAppUnregistered is received by the HMI.

src/js/AppStore.js
iCollin: Good point, I applied these changes here: https://github.com/smartdevicelink/generic_hmi/pull/291/commits/ab9eaef5a91b50c9af0b59a089cf96ae4e7909f5

src/js/Controllers/UIController.js
iCollin: You're right, there's an issue here if an app registers as WEB_VIEW and changes templates. That app's iframe would still be shown. I will look into this.

src/js/Templates/WebView/WebView.js
iCollin: Exactly, the iframe has to be outside the route context so it can run when we are in a menu etc.

src/js/reducers.js
iCollin: updated in https://github.com/smartdevicelink/generic_hmi/pull/291/commits/89d551f877e2ce71bde1d3f7092c3ff225a8acf6

If we changed nonmedia we could fix [this special case](https://github.com/smartdevicelink/generic_hmi/pull/291/files#diff-63ab2bf602f05ad65cee8459f425cc94R52) and just use .toLower() for the registerApp call.

src/js/Controllers/UIController.js
iCollin: In https://github.com/smartdevicelink/generic_hmi/pull/291/commits/a9d1f48c833287bdebb4d097b5e5794e06a7cb57 I added a variable to the state to check whether we were in `/web-view` and used this in index.js alongside this variable that reflected the current context. This worked fine, but I realized that checking if we are in `/web-view` is enough and the context doesn't matter after that. So in https://github.com/smartdevicelink/generic_hmi/pull/291/commits/a1ab7d83384407b8e2808c4727fc9ccfe0bba6e2 I removed the old state variable related to the context.

JackLivio:
FYI style does not persist over ignition cycles for installed projection apps.
Screen Shot 2020-08-04 at 11 11 21 AM

iCollin:
In https://github.com/smartdevicelink/generic_hmi/pull/291/commits/51ed7e7a33cef0e3aa744a992cc9e7aa15bfe459 I removed webengine app style from the state completely, since iframe visibility will now be controlled by the web-view route modifying `webViewActive` in the state.

I also made it so that all web engine apps will be exited when we are running low on memory instead of just projection apps because I was differentiating them by the style.

Lastly I changed when GetInstalledApps is sent to core so that it won't be sent before the HMI is connected to Core.

PR #302, opened 2020-10-09
Process menuIcon and menuTitle properly

jacobkeeler:
- Removes capabilities for menuTitle
- Adds support for menuIcon (replaces the app icon next to the menu when provided)

### DIFF
--- a/src/js/Templates/Shared/Image.js
+++ b/src/js/Templates/Shared/Image.js
@@ -1,13 +1,12 @@
 import React from 'react';
 import store from '../../store'
-import path from 'path'
 import UIController from '../../Controllers/UIController'
 import { connect } from 'react-redux';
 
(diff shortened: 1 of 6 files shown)


src/js/Templates/Shared/Image.js
```diff
@@ -83,6 +83,7 @@ class Image extends React.Component {
         // Refresh image after onPutFile in case of error
         if (this.state.error && this.props.refresh) {
             this.setState({error: false});
+            this.props.refresh = false
```
JackLivio, line 86: I think its bad practice to modify a components props in the render function. Is this change necessary? This statement should not execute since state.error is set to false. The app.refresh prop should be set to false in reducers.js on the next state change.

JackLivio (approved):

PR #304, opened 2020-10-13
Wait for core connection before executing fs controller startup

JackLivio:
Wraps the file system controller startup logic in a timed interval waiting for the core connection to be opened. There was a part of this block that already had a timed wait, so i applied the same approach to the entire block.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -121,8 +121,7 @@ class HMIApp extends React.Component {
             }, 10000, this);
         }
 
-        var waitCoreInterval = setInterval(() => {
-            var sdlSocket = this.sdl.socket
+        var waitCoreInterval = setInterval((sdlSocket) => {
             if (sdlSocket.readyState === sdlSocket.OPEN) {
                 setTimeout(() => { // give time to reply to IsReady
                     FileSystemController.connect(flags.FileSystemApiUrl).then(() => {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -121,44 +121,43 @@ class HMIApp extends React.Component {
...
-                        store.dispatch(updateInstalledAppStoreApps(appEntry));
-                        bcController.getAppProperties(app.policyAppID);
-                        return true;
-                    });
-                    return true;
-                });
-            });
-    
+        var waitCoreInterval = setInterval(() => {
             var sdlSocket = this.sdl.socket
```
iCollin, line 125:
```suggestion
        var waitCoreInterval = setInterval((sdlSocket) => {
```

src/index.js
```diff
@@ -121,44 +121,43 @@ class HMIApp extends React.Component {
...
-                    }, 500);
-                    clearInterval(waitCoreInterval);
-                }
-            }, 500);
-        }, () => { store.dispatch(updateAppStoreConnectionStatus(false)); });
+                    }, () => { store.dispatch(updateAppStoreConnectionStatus(false)); });
+                }, 500); // setTimeout
+                clearInterval(waitCoreInterval);
+            }
+        }, 500); // setInterval
```
iCollin, line 160:
```suggestion
        }, 500, this.sdl.socket); // setInterval
```

iCollin (commented):
`this` is accessed inside of the closure atm

iCollin (approved):

PR #306, opened 2020-10-14
Add onSystemContext to cancel interaction calls

JackLivio:
Adds onSystemContext(MAIN) calls to cancel interaction requests.

### DIFF
--- a/src/js/Controllers/UIController.js
+++ b/src/js/Controllers/UIController.js
@@ -21,6 +21,20 @@ import store from '../store'
 import sdlController from './SDLController'
 import SubmenuDeepFind from '../Utils/SubMenuDeepFind'
 
+const getNextSystemContext = () => {
+    const state = store.getState();
+    const activeApp = state.activeApp;
+    const pathName = window.location.pathname;
+    const inMenuContext = (pathName.includes("/inappmenu") 
+        || pathName.includes("/inapplist")) 
+        && activeApp && state.ui[activeApp] 
+        && !state.ui[activeApp].isPerformingInteraction;
+    if (inMenuContext) {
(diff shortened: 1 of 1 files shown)


iCollin:
my concern is what if the HMI was not in MAIN context before the interaction? For ex Perform Interaction or a SubtleAlert in LIMITED while another app is active.

JackLivio:
SystemContext of MAIN still applies to app's that are in HMI_LEVEL LIMITED/BACKGROUND

src/js/Controllers/UIController.js
```diff
@@ -21,6 +21,20 @@ import store from '../store'
 import sdlController from './SDLController'
 import SubmenuDeepFind from '../Utils/SubMenuDeepFind'
 
+const getNextSystemContext = () => {
+    const state = store.getState();
+    const activeApp = state.activeApp;
+    const pathName = window.location.pathname;
```
iCollin:
```suggestion
    const pathName = window.location.hash;
```

you could potentially use === with the hash property

src/js/Controllers/UIController.js
```diff
@@ -21,6 +21,20 @@ import store from '../store'
 import sdlController from './SDLController'
 import SubmenuDeepFind from '../Utils/SubMenuDeepFind'
 
+const getNextSystemContext = () => {
+    const state = store.getState();
+    const activeApp = state.activeApp;
+    const pathName = window.location.href;
```
iCollin, line 27:
```suggestion
    const pathName = window.location.hash;
```

could use `===` with hash property

iCollin (approved):

PR #314, opened 2020-10-20
Fix/Image validation and invalid image response

ShobhitAd:
SDL HMI issue https://github.com/smartdevicelink/sdl_hmi/issues/314 also exists in the generic hmi

This PR implements image validation and returns a WARNINGS response to SDL if the image does not exist

### DIFF
--- a/src/js/Controllers/UIController.js
+++ b/src/js/Controllers/UIController.js
@@ -376,6 +376,8 @@ class UIController {
     }
     onAlertTimeout(msgID, appID, context, isSubtle) {
         delete this.timers[msgID]
+
+        let imageValidationSuccess = GetValidationResult(msgID)
         RemoveImageValidationRequest(msgID)
 
         store.dispatch(closeAlert(
@@ -383,9 +385,9 @@ class UIController {
             appID
         ))
         const rpc = isSubtle
-            ? RpcFactory.SubtleAlertErrorResponse(msgID, 10, 'subtle alert timed out')
+            ? RpcFactory.SubtleAlertResponse(msgID)
             : RpcFactory.AlertResponse(msgID, appID);
-        this.listener.send(rpc)
+        this.listener.send((imageValidationSuccess) ? rpc : RpcFactory.InvalidImageResponse({ id: rpc.id, method: rpc.result.method }))
 
         const systemContext = getNextSystemContext();
         if (appID !== context) {
--- a/src/js/Controllers/UIController.js
+++ b/src/js/Controllers/UIController.js
@@ -20,7 +20,7 @@ import {
 import store from '../store'
 import sdlController from './SDLController'
 import SubmenuDeepFind from '../Utils/SubMenuDeepFind'
-import { ValidateImages, AddImageValidationRequest, RemoveImageValidationRequest, GetValidationResult } from '../Utils/ValidateImages'
+import { ValidateImages, AddImageValidationRequest, RemoveImageValidationResult } from '../Utils/ValidateImages'
 
 const getNextSystemContext = () => {
     const state = store.getState();
@@ -369,7 +369,7 @@ class UIController {
     }
     onPerformInteractionTimeout(msgID, appID) {
         delete this.timers[msgID]
-        RemoveImageValidationRequest(msgID)
+        RemoveImageValidationResult(msgID)
 
         this.listener.send(RpcFactory.UIPerformInteractionTimeout(msgID))
         store.dispatch(timeoutPerformInteraction(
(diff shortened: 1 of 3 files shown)


src/js/Controllers/UIController.js
```diff
@@ -96,7 +97,19 @@ class UIController {
...
+                let showImages = [rpc.params.graphic, rpc.params.secondaryGraphic];
+                if (rpc.params.softButtons) {
+                    rpc.params.softButtons.forEach (softBtn => {
+                        if (softBtn.image) { showImages.push(softBtn.image); }
+                    });
+                }
+
+                ValidateImages(showImages).then(
+                    () => {this.listener.send(RpcFactory.UIShowResponse(rpc))},
+                    () => {this.listener.send(RpcFactory.InvalidImageResponse(rpc))}
```
iCollin, line 110: in the case of an invalid template and invalid images, i think the error should be forwarded but this will forward the warnings.

I would call UIShowResponse first and update the response if code is success or warnings, probably appending info if warnings

src/js/Controllers/UIController.js
```diff
@@ -401,7 +467,13 @@ class UIController {
     onChoiceSelection(choiceID, appID, msgID) {
         clearTimeout(this.timers[msgID])
         delete this.timers[msgID]
-        this.listener.send(RpcFactory.UIPerformInteractionResponse(choiceID, appID, msgID))
+
+        let imageValidationSuccess = GetValidationResult(msgID)
+        RemoveImageValidationRequest(msgID)
+
+        const rpc = (imageValidationSuccess) ? RpcFactory.UIPerformInteractionResponse(choiceID, appID, msgID)
+            : RpcFactory.InvalidImageResponse({ id: msgID, method: "UI.PerformInteraction" });
```
iCollin, line 475: I'm not sure about this one, if an image is failing to load perform interaction won't be able to get a choiceID back. I think the parameter (choiceID) should take precedent over the info (image validation) here. Unfortunately there is no way to have both params and info. What do you think?

src/js/Controllers/UIController.js
```diff
@@ -381,6 +443,9 @@ class UIController {
         clearTimeout(this.timers[alert.msgID])
         delete this.timers[alert.msgID]
 
+        let imageValidationSuccess = GetValidationResult(alert.msgID);
+        RemoveImageValidationRequest(alert.msgID)
```
iCollin, line 447: I feel like these methods should be combined some how, do you think maybe Get should call Remove?

iCollin (commented):
Just to confirm, the generic_hmi doesn't use these and don't need to validate them, correct?
```
SetGlobalProperties: vrHelp
PerformInteraction: vrHelp
```

src/js/Controllers/UIController.js
ShobhitAd: Fixed in https://github.com/smartdevicelink/generic_hmi/pull/314/commits/b3a7ec4950a8d859cb756112ab24d79971945e59

src/js/Controllers/UIController.js
ShobhitAd: Agreed. Made the change in https://github.com/smartdevicelink/generic_hmi/pull/314/commits/b14ac9f32393890425af94b6ab463678f8bc728a. Let  me know if that looks good

ShobhitAd:
> Just to confirm, the generic_hmi doesn't use these and doesn't need to validate them, correct?
> 
> ```
> SetGlobalProperties: vrHelp
> PerformInteraction: vrHelp
> ```

@iCollin Yes, the `vrHelp` param is not used in the `UIController` at the moment so I think we shouldn't validate the images for that param. Let me know what you think.

src/js/Controllers/UIController.js
ShobhitAd: I think you're right. PerformInteraction should still function normally even if the images in the choice set are invalid. 

Since we can't return the choiceId in a warnings response I've removed the image validation for the perform interaction request in https://github.com/smartdevicelink/generic_hmi/pull/314/commits/c0000f7625983cdbecaaa77176254ec3513e0465

src/js/Controllers/UIController.js
```diff
@@ -457,6 +514,7 @@ class UIController {
         for (var msgID in this.timers) {
             clearTimeout(this.timers[msgID])
             delete this.timers[msgID]
+            RemoveImageValidationResult(msgID)
```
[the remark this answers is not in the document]
ShobhitAd: Removed in https://github.com/smartdevicelink/generic_hmi/pull/314/commits/92c8b151a8a942fc91fd660bd8d46f81f7bf8b2c

iCollin (approved):

PR #325, opened 2020-11-23
Add image optimizations

### LINKED ISSUE #293: Tile images broken on Safari
## Bug Report 

When using the Safari web browser (Version 13.1.2), the tile images either 
1. Show up correctly but after clicking on the button, the location of the image shifts to a random location on the screen.
1. Don't show up in the correct location.

I do not see the issue on Google Chrome.

### Reproduce
1. Set the template to `TILES_ONLY` using either `SetDisplayLayout` or `Show` and send a soft button with both an image (the image should be templated) and text.

#### Expected Result
The images show up in the tile and stay in the tile after the button is clicked.

#### Actual Result
The image either does not show up in the tile or it shifts after the button is clicked.

button image shifted

JackLivio:
Fixes #293 

Adds some optimizations to the image component. 

1. Adds shouldComponentUpdate lifecycle method. This should prevent images from unneccesarily updating when no props have changed. This especially helps with template images since it prevents the canvas from being redrawn unnecessarily. 

2. Storing calculated canvas style properties. This point more-so addresses the issue in #293. For some unknown reason on Safari 14, the calculated canvas height kept growing causing images to shift downward. The calculated styles for the canvas are now only calculated once and stored.

### DIFF
--- a/src/js/Templates/Shared/Image.js
+++ b/src/js/Templates/Shared/Image.js
@@ -69,7 +69,6 @@ class Image extends React.Component {
                     this.canvasStyle.x = (canvas.width / 2) - (scaledDimensions.width / 2);
                     this.canvasStyle.y = (canvas.height / 2) - (scaledDimensions.height / 2);
                 }
-                }
 
                 ctx.drawImage(img, this.canvasStyle.x, this.canvasStyle.y, scaledDimensions.width, scaledDimensions.height);
 
--- a/src/js/Templates/Shared/Image.js
+++ b/src/js/Templates/Shared/Image.js
@@ -65,19 +65,13 @@ class Image extends React.Component {
 
                 ctx.clearRect(0, 0, this.canvasStyle.width, this.canvasStyle.height);
 
-                var x, y;
-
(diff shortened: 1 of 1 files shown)


src/js/Templates/Shared/Image.js
```diff
@@ -30,28 +37,46 @@ class Image extends React.Component {
...
+                ctx.clearRect(0, 0, this.canvasStyle.width, this.canvasStyle.height);
 
-                var x = (canvas.width / 2) - (scaledDimensions.width / 2);
-                var y = (canvas.height / 2) - (scaledDimensions.height / 2);
-                ctx.drawImage(img, x, y, scaledDimensions.width, scaledDimensions.height);
+                if (!this.canvasStyle.x || !this.canvasStyle.y) {
+                    this.canvasStyle.x = (canvas.width / 2) - (scaledDimensions.width / 2);
+                    this.canvasStyle.y = (canvas.height / 2) - (scaledDimensions.height / 2);
+                }
+                }
```
iCollin, line 72:
```suggestion
                }
```

iCollin (approved):

PR #327, opened 2020-12-02
Ignore GetVehicleData

JackLivio:
Causing issues with manticore because the Generic HMI was immediately responding with Generic Error. This change makes the generic hmi ignore the request so the vehicle data component of manticore can respond instead.


jacobkeeler (changes requested):

JackLivio:
Closing since manticore adopted their own fix for this issue.

PR #328, opened 2020-12-02
Fix crash with VR_ONLY PerformInteraction request

jacobkeeler:
Send INVALID_DATA response if choiceSet is not provided and add safety check for app.choices

### DIFF
--- a/src/js/Controllers/RpcFactory.js
+++ b/src/js/Controllers/RpcFactory.js
@@ -23,6 +23,19 @@ class RpcFactory {
         }
         return msg;
     }
+    static ErrorResponse(rpc, code, info) {
+        return ({
+            "jsonrpc": "2.0",
+            "id": rpc.id,
+            "error": {
+                "code": code,
+                "message": info,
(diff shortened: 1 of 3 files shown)


src/js/Controllers/RpcFactory.js
```diff
@@ -23,6 +23,19 @@ class RpcFactory {
         }
         return msg;
     }
+    static ErrorResponse(rpc, code, info) {
```
ShobhitAd, line 26: Might be a better idea to modify the existing `UIPerformInteractionFailure` method.

ShobhitAd (changes requested):

src/js/Controllers/RpcFactory.js
jacobkeeler: Eh, I find it a bit more confusing that there isn't a more generic error response method. If you want, I could modify that method to use this one, but I don't think this should be removed

ShobhitAd (approved):

PR #336, opened 2020-12-17
restrict alert sb image max size

### LINKED ISSUE #323: Soft button content is misaligned in an alert
## Bug Report
If a soft button in an alert has both images and text, the text extends outside the bounds of the button. This was seen on Firefox web browser (Mozilla Firefox 83.0)

### Reproduction Steps
1. Create an alert with two soft buttons. One button is text-only ("Ok") and the second button has text and an image ("Gotcha" with the speech bubble icon). 
1. Send the alert request. 

#### Expected Result
The soft button text and image is center aligned inside the button.

#### Actual Result
The soft button text and image are misaligned. Text extends outside the bounds of the soft button. 

Screen Shot 2020-11-18 at 2 05 46 PM

iCollin:
Fixes #323

### DIFF
--- a/src/css/components/_alert.scss
+++ b/src/css/components/_alert.scss
@@ -131,6 +131,10 @@
     @include justify-content(center);
 }
 
+.btn-type-BOTH .soft-button-image {
+    max-width: 50%;
+}
+
 .alert-button-3 {
     //height: 123.333333334px;
     height: 75px;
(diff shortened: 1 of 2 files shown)


src/css/components/_alert.scss
```diff
@@ -131,6 +131,10 @@
     @include justify-content(center);
 }
 
+.btn-type-BOTH .soft-button-image {
+    max-width: 50%;
+}
```
jacobkeeler, line 136: Unfortunately I'm seeing similarly odd things in other cases with BOTH softbuttons. Notably it seems to occur with templated dynamic images specifically

jacobkeeler (changes requested):

iCollin:
Closed in favor of https://github.com/smartdevicelink/generic_hmi/pull/338

PR #345, opened 2021-01-12
Feature/video streaming

iCollin:
Adds Video Streaming
Fixes https://github.com/smartdevicelink/generic_hmi/issues/290
Fixes https://github.com/smartdevicelink/generic_hmi/issues/315


tools/start_server.py
```diff
@@ -43,16 +43,20 @@
 from http.server import SimpleHTTPRequestHandler
```
iCollin, line 43: i accidentally made these whitespace changes, please let me know if I should revert them

tools/start_server.py
```diff
@@ -174,12 +178,13 @@ def __init__(self, _websocket, _path):
       "InstallApp": RPCService.webengine_manager.handle_install_app,
       "GetInstalledApps": RPCService.webengine_manager.handle_get_installed_apps,
       "UninstallApp": RPCService.webengine_manager.handle_uninstall_app,
+      "StartVideoStream": self.handle_start_video_stream,
```
JackLivio, line 181: I think there needs to be a message added for Stopping a video stream.

src/index.js
```diff
@@ -70,8 +78,85 @@ class HMIApp extends React.Component {
...
+            return;
+        }
+
+        this.touchInProgress = false;
+        this.onTouchEvent('END', event);
+    }
+
     render() {
         const themeClass = this.state.dark ? 'dark-theme' : 'light-theme';
+        var videoStyle = { position: 'absolute', width: 960, height: 600, top: 75, left: 0, backgroundColor: 'transparent', objectFit: 'fill' };
```
JackLivio, line 145: Might want to try importing the master height / width and setting video style that way.

src/js/Controllers/DisplayCapabilities.js
```diff
@@ -544,6 +544,65 @@ let capabilities = {
 			"driverDistractionCapability": {
 				"menuLength": 3,
 				"subMenuDepth": 2
+			},
+			videoStreamingCapability: {
+				preferredResolution: {
+					resolutionWidth: 960,
```
JackLivio, line 550: if the scss master-height/width variables can be imported into react, it would be cool if those config settings would determine the supported resolutions.

src/index.js
JackLivio, line 145: Streaming in any template is interesting. I dont mind it, but I think we should add a basic Full Screen Nav Map as a supported template. It can just be an empty page with the header at the top. If we have time we can add the nav buttons + text.

src/js/Controllers/DisplayCapabilities.js
iCollin: done in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/6125f5abb20538a97e15a641fabaa88dc9a1c3de

src/index.js
iCollin: done in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/6125f5abb20538a97e15a641fabaa88dc9a1c3de and also subtracted the appheader height in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/97519be979874f42cb47739822067a6a536c4bea

iCollin: Good point, we definitely need a NAV template, but the actual video DOM element needs to be in this component so that it won't be removed when the route changes and the template component is unmounted.

I have added this template in commit 8dbd4bb

src/js/Controllers/BCController.js
```diff
@@ -98,6 +101,18 @@ class BCController {
...
                 this.listener.send(RpcFactory.GetSystemTime(rpc.id))
                 return null
+
+            case "OnAppCapabilityUpdated":
+                if (rpc.params.appCapability.appCapabilityType === 'VIDEO_STREAMING') {
+                    var vsc = rpc.params.appCapability.videoStreamingCapability;
+                    var caps = [ vsc ];
+                    if (vsc.additionalVideoStreamingCapabilities) {
+                        caps.concat(vsc.additionalVideoStreamingCapabilities);
+                        delete vsc.additionalVideoStreamingCapabilities;
```
JackLivio, line 111: Do you mean to delete caps.additionalVideoStreamingCapabilities? Deleting this is modifying the original rpc message and causing the console logs to show that additionalVideoStreamingCapabilities is never received.

JackLivio: Also caps seems to be losing track of the additionalVideoStreaminCapabilities param
Screen Shot 2021-02-01 at 5 20 44 PM

JackLivio: I stringified the request to see if the additional video params were in the request and it was. Last printed line is caps which is also missing the additional params

JackLivio: Oh yea this is probably why

> The concat() method is used to merge two or more arrays. This method does not change the existing arrays, but instead returns a new array.

iCollin: fixed in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/7d647f7bc4fd20a4ddae2717d968b2032819d4bc

tools/start_server.py
```diff
@@ -408,16 +442,20 @@ def uninstall_app(self, _app_id):
 def main():
   parser =  argparse.ArgumentParser(description="Handle backend operations for the hmi")
   parser.add_argument('--host', type=str, required=True, help="Backend server hostname")
-  parser.add_argument('--port', type=int, required=True, help="Backend server port number")
+  parser.add_argument('--ws-port', type=int, required=True, help="Backend server port number")
+  parser.add_argument('--video-port', type=int, default=8085, help="Video streaming server port number")
+  parser.add_argument('--audio-port', type=int, default=8086, help="Video streaming server port number")
```
iCollin, line 447:
```suggestion
  parser.add_argument('--audio-port', type=int, default=8086, help="Audio streaming server port number")
```

src/js/Controllers/BCController.js
```diff
@@ -106,12 +106,7 @@ class BCController {
             case "OnAppCapabilityUpdated":
                 if (rpc.params.appCapability.appCapabilityType === 'VIDEO_STREAMING') {
                     var vsc = rpc.params.appCapability.videoStreamingCapability;
-                    var caps = [ Object.assign({}, vsc) ];
-                    delete caps[0].additionalVideoStreamingCapabilities;
-                    if (vsc.additionalVideoStreamingCapabilities) {
-                        caps = caps.concat(vsc.additionalVideoStreamingCapabilities);
-                    }
-                    store.dispatch(setVideoStreamingCapability(rpc.params.appID, caps));
+                    store.dispatch(setVideoStreamingCapability(rpc.params.appID, vsc.additionalVideoStreamingCapabilities));
```
JackLivio, line 109: Check if additionalVideoStreamingCapabilities exists and has a length > 0, if not pass empty array?

src/js/Controllers/DisplayCapabilities.js
```diff
@@ -603,6 +603,25 @@ let capabilities = {
 				scale: 1,
 				preferredFPS: 30,
 				additionalVideoStreamingCapabilities: [
+					{
```
JackLivio, line 606: My understanding of the conversation with mobile teams was that this was not necessary. Mobile will parse the top level.

src/js/Controllers/NavController.js
```diff
@@ -73,6 +75,13 @@ class NavController {
                 this.audioStreamUrl = rpc.params.url;
                 return {"rpc": RpcFactory.StartAudioStreamSuccess(rpc.id)};
             case "SetVideoConfig":
+                this.videoConfig = rpc.params.config;
+
+                if (this.webmSupport === undefined) {
+                    var video = document.getElementById('navi_stream');
+                    this.webmSupport = 'probably' === video.canPlayType('video/webm; codecs="vp8"');
```
JackLivio, line 84: This returns 'probably'?

iCollin: probably, maybe, or an empty string if it knows it does not support

tools/start_server.py
```diff
@@ -273,15 +273,32 @@ def handle_start_video_stream(self, _method, _params):
...
-
-    if index != 0:
-      return self.gen_error_msg('Streaming data not available from SDL')
-
-    return { 'success': True, 'params': { 'endpoint': server_endpoint } }
+    if 'config' not in _params:
+      return self.gen_error_msg('Missing mandatory param \'config\'')
+
+    if _params['config']['protocol'] == 'RTP':
+      print('\033[33mFFMpeg does not support RTP video\033[0m')
```
JackLivio, line 280: Say: SDL does not support in browser RTP video

FFMPEG says it supports RTP, we just cant get it to work :(

src/js/Controllers/BCController.js
iCollin: passing an empty array in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/37bf9edd9720d6105971200fa1fabfb3164d97b8

I also made it so the resolution drop down only shows if there are valid options

tools/start_server.py
iCollin: right, its RTP over TCP it doesn't support. comment updated in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/14fa5705a549a868d956d456fe3c68d51f1ed14f

src/js/Controllers/DisplayCapabilities.js
iCollin: I agree, removed duplicate capability in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/82eb1a16f2739e5e2bccd7755c02278356cab691

src/js/Controllers/BCController.js
```diff
@@ -98,6 +102,16 @@ class BCController {
             case "GetSystemTime":
                 this.listener.send(RpcFactory.GetSystemTime(rpc.id))
                 return null
+
+            case "OnAppCapabilityUpdated":
+                if (rpc.params.appCapability.appCapabilityType === 'VIDEO_STREAMING') {
+                    var vsc = rpc.params.appCapability.videoStreamingCapability;
+                    if (!vsc.additionalVideoStreamingCapabilities) {
```
JackLivio, line 109: also check if vsc is valid

iCollin: done in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/81b4a86651d856a228ef2179a99da9dfc5b269d6

JackLivio (approved):

tools/start_server.py
```diff
@@ -263,14 +269,59 @@ def isFileNameValid(_file_name):
...
+
+    if _params['config']['protocol'] == 'RTP':
+      print('\033[33mSDL does not support RTP video in browser\033[0m')
+      if _params['config']['codec'] == 'H264':
+        print('\033[1mYou may view your video with gstreamer:\033[0m')
+        print('gst-launch-1.0 souphttpsrc location=' + _params['url'] + ' ! "application/x-rtp-stream" ! rtpstreamdepay ! "application/x-rtp,media=(string)video,clock-rate=90000,encoding-name=(string)H264" ! rtph264depay ! "video/x-h264, stream-format=(string)avc, alignment=(string)au" ! avdec_h264 ! videoconvert ! ximagesink sync=false')
+
+    elif not _params['webm']:
+      print('\033[33mYour browser does not support WEBM video\033[0m')
+      if _params['config']['protocol'] == 'RAW' and _params['config']['protocol'] == 'H264':
```
AKalinich-Luxoft, line 287:
```suggestion
      if _params['config']['protocol'] == 'RAW' and _params['config']['codec'] == 'H264':
```

tools/start_server.py
```diff
@@ -263,14 +269,59 @@ def isFileNameValid(_file_name):
...
     }
 
+  def handle_start_video_stream(self, _method, _params):
+    if 'url' not in _params:
+      return self.gen_error_msg('Missing mandatory param \'url\'')
+
+    if 'config' not in _params:
+      return self.gen_error_msg('Missing mandatory param \'config\'')
+
+    if _params['config']['protocol'] == 'RTP':
```
AKalinich-Luxoft, line 279: I would recommend to use `get()` instead of direct map access because all config parameters are not mandatory. Python will throw the exception if you try to access key which does not exist

tools/start_server.py
iCollin: applied in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/ba8ca359b09050171923c8073e10909db86dd9e6

tools/start_server.py
iCollin: good idea, non mandatory access replaced with get in https://github.com/smartdevicelink/generic_hmi/pull/345/commits/ba8ca359b09050171923c8073e10909db86dd9e6

JackLivio (approved):

PR #349, opened 2021-01-19
Add timer offset logic to handle RESUME event

jacobkeeler:
Fixes issue with #329 

### Testing Plan

#### Core Tests
- Sending multiple PAUSE and RESUME events with different playback rates
- Sending multiple RESUME events in a row with different playback rates

Core version / branch / commit hash / module tested against: develop
Proxy+Test App name / version / branch / commit hash / module tested against: RPC Builder App JS, master

### Summary
Adds proper logic for resuming timer by storing the timer offset when a pause/resume event is received

##### Bug Fixes
* Fixes SetMediaClockTimer behavior when a RESUME event is received with a specific countRate

### CLA

### DIFF
--- a/src/js/reducers.js
+++ b/src/js/reducers.js
@@ -33,7 +33,8 @@ function newAppState () {
         audioStreamingIndicator: "PLAY_PAUSE",
         countRate: 1.0,
         updateTime: new Date().getTime(),
-        pauseTime: null,
+        timerOffset: null,
+        paused: false,
         isDisconnected: false,
         displayLayout:  null,
         alert: {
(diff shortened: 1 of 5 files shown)


src/js/reducers.js
```diff
@@ -447,38 +448,45 @@ function ui(state = {}, action) {
...
                 app.updateTime = new Date().getTime()
+                app.timerOffset = null
             }
             else if (action.updateMode === "COUNTDOWN") {
                 if (action.updateMode !== app.countDirection) {
                     app.endTime = action.endTime ? action.endTime : null
                 }
                 app.countDirection = action.updateMode
                 app.updateTime = new Date().getTime()
+                app.timerOffset = null
```
ShobhitAd, line 459: Wouldn't it be better to set the value to 0(instead of null) and directly use `app.timerOffset` in `MediaTrackInfo_c.js` and `ProgressBar_c.js`

ShobhitAd (changes requested):

ShobhitAd (approved):

JulianKast (approved):
Works with Java_Suite

PR #373, opened 2021-03-26
handle case image is error and then path changes

iCollin:
Fixes https://github.com/smartdevicelink/generic_hmi/issues/377

### Summary
if you set an image to a path that doesn't exist, and do not putfile it, and then change path to a valid image, the image would not previously render. this PR makes it so that when the path changes the error flag is reset

### CLA

### DIFF
--- a/src/js/Templates/Shared/Image.js
+++ b/src/js/Templates/Shared/Image.js
@@ -6,7 +6,7 @@ import { connect } from 'react-redux';
 class Image extends React.Component {
     constructor(props) {
         super(props);
-        this.state = {error: false, refreshed: false};
+        this.state = {errorPath: null, refreshed: false};
         this.initialCanvasDimensions = {
             height: null,
             width: null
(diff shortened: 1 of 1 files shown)


src/js/Templates/Shared/Image.js
```diff
@@ -129,6 +129,11 @@ class Image extends React.Component {
         else if (!prevState.refreshed) {
             newState.error = false
         }
+        // If there was an error loading image but image path has changed,
+        // remove error flag
+        if (prevState.error && prevState.error !== nextProps.image) {
+            newState.error = false;
```
jacobkeeler, line 135:
```suggestion
            newState.errorPath = null;
```
Just so we're not mixing data types for this value, perhaps we should rename it

jacobkeeler (changes requested):

src/js/Templates/Shared/Image.js
iCollin: Good idea, I updated all uses of state.error in https://github.com/smartdevicelink/generic_hmi/pull/373/commits/aaea83e1eb0cda14677c3366d3cb3e73ffee3543

PR #400, opened 2021-06-17
cancel interaction only applies to interactions from same app

iCollin:
### Summary
Change CancelInteraction code so that an app may only cancel it's own interactions and not those of other apps.

#### Testing Plan
Cancel an interaction successfully
Try to cancel an interaction of another app (should be ignored)

### CLA

### DIFF
--- a/src/js/Controllers/UIController.js
+++ b/src/js/Controllers/UIController.js
@@ -419,60 +419,59 @@ class UIController {
 
                 return null
             }
-            case "CancelInteraction":
+            case "CancelInteraction": {
 
-                const state2 = store.getState()
-                for(const appID in state2.ui) {
-                    const app = state2.ui[appID];
+                const state = store.getState();
+                const app = state.ui[rpc.params.appID];
 
-                    if (rpc.params.functionID === 10 && app.isPerformingInteraction
-                         && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.interactionCancelId)) {
-                        clearTimeout(this.timers[app.interactionId])
-                        delete this.timers[app.interactionId]
-                        this.listener.send(RpcFactory.UIPerformInteractionCancelledResponse(app.interactionId))
-                        store.dispatch(deactivateInteraction(rpc.params.appID))
-                        this.onSystemContext("MAIN", rpc.params.appID)
-                        return true
-                    } else if (rpc.params.functionID === 12 && app.alert.showAlert && !app.alert.isSubtle
-                         && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.alert.cancelID)) {
-                        clearTimeout(this.timers[app.alert.msgID])
-                        delete this.timers[app.alert.msgID]
-                        this.listener.send(RpcFactory.AlertAbortedResponse(app.alert.msgID))
-                        store.dispatch(closeAlert(app.alert.msgID, rpc.params.appID))
-                        const context = getNextSystemContext();
-                        this.onSystemContext(context, rpc.params.appID)
-                        return true
-                    } else if (rpc.params.functionID === 64 && app.alert.showAlert && app.alert.isSubtle
+                if (rpc.params.functionID === 10 && app.isPerformingInteraction
+                        && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.interactionCancelId)) {
+                    clearTimeout(this.timers[app.interactionId])
+                    delete this.timers[app.interactionId]
+                    this.listener.send(RpcFactory.UIPerformInteractionCancelledResponse(app.interactionId))
+                    store.dispatch(deactivateInteraction(rpc.params.appID))
+                    this.onSystemContext("MAIN", rpc.params.appID)
+                    return true
+                } else if (rpc.params.functionID === 12 && app.alert.showAlert && !app.alert.isSubtle
                         && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.alert.cancelID)) {
-                       clearTimeout(this.timers[app.alert.msgID])
-                       delete this.timers[app.alert.msgID]
-                       this.listener.send(RpcFactory.SubtleAlertErrorResponse(app.alert.msgID, 5, 'subtle alert was cancelled'))
-                       store.dispatch(closeAlert(app.alert.msgID, rpc.params.appID))
-                       const context = getNextSystemContext();
-                       this.onSystemContext(context, rpc.params.appID)
-                       return true
-                    } else if (rpc.params.functionID === 26 && app.slider.showSlider
-                        && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.slider.cancelID)) {
-                       clearTimeout(this.timers[app.slider.msgID])
-                       delete this.timers[app.slider.msgID]
-                       this.listener.send(RpcFactory.SliderAbortedResponse(app.slider.msgID))
-                       store.dispatch(closeSlider(app.alert.msgID, rpc.params.appID))
-                       const context = getNextSystemContext();
-                       this.onSystemContext(context, rpc.params.appID)
-                       return true
-                    } else if (rpc.params.functionID === 25 && app.scrollableMessage.active
-                        && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.scrollableMessage.cancelID)) {
-                      clearTimeout(this.timers[app.scrollableMessage.msgID]);
-                      delete this.timers[app.scrollableMessage.msgID];
-                      this.listener.send(RpcFactory.ScrollableMessageAbortedResponse(app.scrollableMessage.msgID));
-                      store.dispatch(closeScrollableMessage(app.alert.msgID, rpc.params.appID));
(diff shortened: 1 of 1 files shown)


src/js/Controllers/UIController.js
```diff
@@ -419,60 +419,59 @@ class UIController {
...
+                if (rpc.params.functionID === 10 && app.isPerformingInteraction
+                        && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.interactionCancelId)) {
+                    clearTimeout(this.timers[app.interactionId])
+                    delete this.timers[app.interactionId]
+                    this.listener.send(RpcFactory.UIPerformInteractionCancelledResponse(app.interactionId))
+                    store.dispatch(deactivateInteraction(rpc.params.appID))
+                    this.onSystemContext("MAIN", rpc.params.appID)
+                    return true
+                } else if (rpc.params.functionID === 12 && app.alert.showAlert && !app.alert.isSubtle
                         && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.alert.cancelID)) {
```
jacobkeeler, line 436:
```suggestion
                    && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.alert.cancelID)) {
```

jacobkeeler (changes requested):

src/js/Controllers/UIController.js
```diff
@@ -419,60 +419,59 @@ class UIController {
...
+                    this.onSystemContext(context, rpc.params.appID)
+                    return true
+                } else if (rpc.params.functionID === 25 && app.scrollableMessage.active
+                    && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.scrollableMessage.cancelID)) {
+                    clearTimeout(this.timers[app.scrollableMessage.msgID]);
+                    delete this.timers[app.scrollableMessage.msgID];
+                    this.listener.send(RpcFactory.ScrollableMessageAbortedResponse(app.scrollableMessage.msgID));
+                    store.dispatch(closeScrollableMessage(app.alert.msgID, rpc.params.appID));
+                    const context = getNextSystemContext();
+                    this.onSystemContext(context, state.activeApp);
```
jacobkeeler, line 469:
```suggestion
                    this.onSystemContext(context, rpc.params.appID);
```

jacobkeeler (changes requested):

src/js/Controllers/UIController.js
iCollin: I don't understand this, the app that is in full may have its context changed, not the App which sent CancelInteraction. It may have put an alert over another app. If sending app is in background then it's context isn't changed. Can you help me understand this suggestion?

jacobkeeler: Hmmm, can't imagine a ScrollableMessage being shown from the background, but if we're accounting for that we'll probably need to change a few of these other cases as well.

src/js/Controllers/UIController.js
```diff
@@ -490,60 +490,59 @@ class UIController {
...
+                    this.onSystemContext("MAIN", rpc.params.appID)
+                    return true
+                } else if (rpc.params.functionID === 12 && app.alert.showAlert && !app.alert.isSubtle
+                    && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.alert.cancelID)) {
+                    clearTimeout(this.timers[app.alert.msgID])
+                    delete this.timers[app.alert.msgID]
+                    this.listener.send(RpcFactory.AlertAbortedResponse(app.alert.msgID))
+                    store.dispatch(closeAlert(app.alert.msgID, rpc.params.appID))
+                    const context = getNextSystemContext();
+                    this.onSystemContext(context, rpc.params.appID)
```
jacobkeeler, line 513:
```suggestion
                    if (rpc.params.appID !== state.activeApp) {
                        this.onSystemContext("MAIN", rpc.params.appID)
                    }
                    this.onSystemContext(context, state.activeApp)
```

src/js/Controllers/UIController.js
```diff
@@ -490,60 +490,59 @@ class UIController {
...
+                    return true
+                } else if (rpc.params.functionID === 64 && app.alert.showAlert && app.alert.isSubtle
+                    && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.alert.cancelID)) {
+                    clearTimeout(this.timers[app.alert.msgID])
+                    delete this.timers[app.alert.msgID]
+                    this.listener.send(RpcFactory.SubtleAlertErrorResponse(app.alert.msgID, 5, 'subtle alert was cancelled'))
+                    store.dispatch(closeAlert(app.alert.msgID, rpc.params.appID))
+                    const context = getNextSystemContext();
+                    this.onSystemContext(context, rpc.params.appID)
```
jacobkeeler, line 522:
```suggestion
                    if (rpc.params.appID !== state.activeApp) {
                        this.onSystemContext("MAIN", rpc.params.appID)
                    }
                    this.onSystemContext(context, state.activeApp)
```

src/js/Controllers/UIController.js
```diff
@@ -490,60 +490,59 @@ class UIController {
...
+                    return true
+                } else if (rpc.params.functionID === 26 && app.slider.showSlider
+                    && (rpc.params.cancelID === undefined || rpc.params.cancelID === app.slider.cancelID)) {
+                    clearTimeout(this.timers[app.slider.msgID])
+                    delete this.timers[app.slider.msgID]
+                    this.listener.send(RpcFactory.SliderAbortedResponse(app.slider.msgID))
+                    store.dispatch(closeSlider(app.alert.msgID, rpc.params.appID))
+                    const context = getNextSystemContext();
+                    this.onSystemContext(context, rpc.params.appID)
```
jacobkeeler, line 531:
```suggestion
                    this.onSystemContext(context, state.activeApp)
```
If we're accounting for this message being sent in the background

jacobkeeler (changes requested):
Also will probably need to apply some of these changes to UI.ClosePopUp to match

src/js/Controllers/UIController.js
iCollin: The steal focus action always makes me consider the case of a background interaction.

I applied your suggestions, and double checked all the context changes sent from within this CancelInteraction handler are now the inverse of those applied when each interaction is first received.

src/js/Controllers/UIController.js
```diff
@@ -490,60 +490,65 @@ class UIController {
...
+                    this.listener.send(RpcFactory.ScrollableMessageAbortedResponse(app.scrollableMessage.msgID));
+                    store.dispatch(closeScrollableMessage(app.alert.msgID, rpc.params.appID));
+                    const context = getNextSystemContext();
+                    this.onSystemContext(context, state.activeApp);
+                    return true;
                 }
                 
                 return { rpc: RpcFactory.UICancelInteractionIgnoredResponse(rpc) }
+            }
             case "ClosePopUp": {
```
jacobkeeler, line 552: We'll probably need to edit the `OnSystemContext` usages here too

jacobkeeler (changes requested):

iCollin:
In https://github.com/smartdevicelink/generic_hmi/pull/400/commits/cd3042a13f7d682d23b495bf8c7a07f81f44696a I aligned the context changes from the OnClose methods with these, ensuring we won't stick a background application in MENU context.

src/js/Controllers/UIController.js
iCollin: Updated the calls from ClosePopUp in https://github.com/smartdevicelink/generic_hmi/pull/400/commits/d5533e893f784e9e88530620ac3df2c7ba27f089

iCollin:
`onDefaultAction` and `onStealFocus ` were both meant to be called with the context parameter being the activeApp, but were sometimes called with the sending app as the context, so I have updated how those methods find the context in https://github.com/smartdevicelink/generic_hmi/pull/400/commits/05c893050cc89245daa0160791dfc770e324abb1 and https://github.com/smartdevicelink/generic_hmi/pull/400/commits/d91a612ee42bd681600636095efd7a6ac9619614.

I also modified Slider to send `HMI_OBSCURED` even if the sending app is in FULL in https://github.com/smartdevicelink/generic_hmi/pull/400/commits/05c893050cc89245daa0160791dfc770e324abb1.

jacobkeeler (changes requested):
```
        <element name="HMI_OBSCURED" internal_name="SYSCTXT_HMI_OBSCURED">
            <description>The app's display HMI is currently being obscured by either a system or other app's overlay.</description>
        </element>
```
Seems that this context is meant to be used when app is obscured by a dialog that is not its own

iCollin:
> Seems that this context is meant to be used when app is obscured by a dialog that is not its own

Interesting, thank you for pointing that out! I have reverted my change to HMI_OBSCURED for Slider and aligned ScrollableMessage to this new information as well.

PerformInteraction is still sending `HMI_OBSCURED` to the requesting app in both of our HMIs, so I did not mess with the context change here.

jacobkeeler (changes requested):

PR #406, opened 2021-06-21
Consent & GetUserFriendlyMessage

iCollin:
### Summary
This PR is ready for review.

This PR adds the following functionality:
- Consent Prompts
- App Permissions View
- Support for GetUserFriendlyMessage RPC
- Non-intrusive Notifications
- Imports Bootstrap (used for switches on app permissions view)
- Support for OnAppPermissionChanged notification
- App Header Custom JSX Icons

This PR also refactors the TTS controller a bit so that the HMI can easily queue TTS and to prevent TTS conflicts.

### To Do
- [x] Find better icon for Header button that takes you to permissions view
- [x] Add mechanism to replace %% variables in CFMs, globally where possible

### Future Considerations
- External Consent Status
- Ability to change HMI Language
- [Burger menu](https://github.com/negomi/react-burger-menu) with Permissions and App Store within it (can then be expanded later for other settings)

### CLA


src/js/Controllers/SDLController.js
```diff
@@ -22,10 +29,42 @@ class SDLController {
...
+                    this.statusMessages['UPDATING'] = 'Updating';
+                    this.statusMessages['UP_TO_DATE'] = 'Update Complete';
+                }
+                
+
+                this.toastStatus(status);
+            });
+        } else {
+            if (this.statusMessages[status].tts) { ttsController.queueTTS(this.statusMessages[status].tts); }
+            toast((_toast) => (<PermissionsPopup _toast={_toast} header={this.statusMessages[status].line1}/>), { duration: 2000 });
```
JackLivio, line 65:
```suggestion
            toast((_toast) => (<PermissionsPopup _toast={_toast} header={this.statusMessages[status].line1}/>), { duration: 5000 });
```


2000 seemed a little too fast

JackLivio:
Hide menu icon when viewing permissions list. Clicking menu causes error
Screen Shot 2021-06-23 at 5 11 15 PM

JackLivio:
I think the data consent prompt for SDL should come before acknowledging the app permissions being requested

JackLivio:
Disabling an allowed permission and then pressing save does not save the state. No messages are sent to Core for disallowing the permission

JackLivio:
Activating an app for the first time is requiring the user to activate the app twice. The data-consent + app permissions view should be chained together with one click to activate an app.

src/js/AppPermissions.js
```diff
@@ -5,19 +5,21 @@ import AppHeader from './containers/Header';
...
+    var pendingActivation = state.system.permissionsAppAwaitingActivation;
 
     return {
         theme: theme,
         editingAppID: editingAppID,
         allowedFunctions: allowedFunctions,
         activeLayout: state.ui[editingAppID]?.displayLayout,
-        appOptions: state.appList
+        appOptions: state.appList,
+        pendingActivation: pendingActivation
```
iCollin, line 22: I realized after writing this that theoretically there can be multiple apps pending activation here, and this should be an array. let me know if you like this solution and if so I will change this variable to handle multiple apps.

src/js/AppPermissions.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    savePermissions() {
+        var allowedFunctions = this.state.allowedFunctions.map((f) => ({
+            id: f.id, name: f.name, allowed: f.allowed
+        }));
+        sdlController.onAppPermissionConsent(this.props.editingAppID, allowedFunctions);
+
+        if (this.props.pendingActivation === this.props.editingAppID) {
+            sdlController.onAppActivated(this.props.editingAppID);
```
JackLivio, line 56: Remove this call to onAppActivated. SDL Core sends an activate app request to hmi after data is consented

JackLivio:
Experiencing an issue with external policies connected with the sdl server.

After activating an app for the first time i am presented with the consent prompt. I agree and then the PTU takes place. After the PTU is successful I am given the list of permissions to accept or deny. After I hit save the view takes me back to the app list instead of the app in full. I am not able to activate the app at this point, the app list is stuck

iCollin:
> Experiencing an issue with external policies connected with the sdl server.
> 
> After activating an app for the first time i am presented with the consent prompt. I agree and then the PTU takes place. After the PTU is successful I am given the list of permissions to accept or deny. After I hit save the view takes me back to the app list instead of the app in full. I am not able to activate the app at this point, the app list is stuck

I was able to reproduce this scenario but in my tests, after the PTU takes place the HMI receives a ChangeRegistration and an ActivateApp with `level=background`. Then the OnAppPermissionChanged notification is received, meaning by the time the permissions view is opened, the app is no longer in FULL. Therefore I expect that when you save the permissions you go back to the app list.

In the current state, after saving permissions, I could not find any issues with activating the app.

src/js/Controllers/SDLController.js
iCollin: They're so quick to digest, and often all three come within 2 seconds. Having all three popups covering half of my screen for 3seconds would annoy me, even knowing this won't happen often at all.

I am not opposed to extending the time a bit, but I think 5s is long, that is what I originally used.

src/js/Controllers/TTSController.js
```diff
@@ -135,33 +70,61 @@ class TTSController {
...
-        this.speakID = null;
-        if (this.currentlyPlaying === "FILE") {
-            this.audioPlayer.onended = null;
-            this.audioPlayer.pause();
-            this.audioPlayer.src = "";
-        } else if (this.currentlyPlaying === "TEXT") {
-            window.speechSynthesis.pause();
-            window.speechSynthesis.cancel();
+        var file = null;
+        while (file = this.filePlaylist.shift()) {
```
JackLivio, line 74: This is causing a warning when building the app

src/js/Controllers/TTSController.js
```diff
@@ -182,26 +145,19 @@ class TTSController {
...
                     };
                 }
                 var ttsChunks = rpc.params.ttsChunks
-                this.filePlaylist = []
                 for (var i=0; i<ttsChunks.length; i++) {
                         this.filePlaylist.push(ttsChunks[i])
                 }
-                this.speakID = rpc.id;
-                this.listener.send(RpcFactory.TTSStartedNotification());
+                this.filePlaylist.push({ type: 'REPLY', id: rpc.id });
```
JackLivio, line 157: Please add code comments for this reply type since a developer might assume this is a type associated with the hmi api

src/js/Controllers/TTSController.js
iCollin: fixed warning in https://github.com/smartdevicelink/generic_hmi/pull/406/commits/d15c407092b3d39f0bdb0869f176896fe7a46b7e

src/js/Controllers/TTSController.js
iCollin: good point, I added a comment in two places in https://github.com/smartdevicelink/generic_hmi/pull/406/commits/4c4152ada070e01fa0501a5b95929969699b115f

JackLivio (approved):

PR #415, opened 2021-07-21
Fix UI GetCapabilities Response

iCollin:
### Summary
reduce scrollableMessageBody row count according to HMI API TextField.rows.maxvalue

The current UI.GetCapabilities response is considered invalid and tossed by Core.
https://github.com/smartdevicelink/sdl_core/blob/develop/src/components/interfaces/HMI_API.xml#L2737

```
<struct name="TextField">
  ...
  <param name="rows" type="Integer" minvalue="1" maxvalue="8" mandatory="true">
    <description>The number of rows of this field.</description>
  </param>
</struct>
```

### CLA

### DIFF
--- a/src/js/Controllers/DisplayCapabilities.js
+++ b/src/js/Controllers/DisplayCapabilities.js
@@ -70,7 +70,7 @@ let textWithGraphicCapabilities = {
 			textField("mainField3"),
 			textField("mainField4"),
 			textField("templateTitle", 50),
-			textField("scrollableMessageBody", 44, 15),
+			textField("scrollableMessageBody", 44, 8),
 			textField("alertText1"),
 			textField("alertText2"),
 			textField("alertText3"),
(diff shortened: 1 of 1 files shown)


src/js/Controllers/DisplayCapabilities.js
```diff
@@ -421,7 +421,7 @@ let capabilities = {
 			"displayType": "SDL_GENERIC",
 			"displayName": "GENERIC_DISPLAY",
 			"textFields": [
-				textField("scrollableMessageBody", 44, 15),
+				textField("scrollableMessageBody", 44, 8),
```
jacobkeeler, line 424: While we are fixing these, I noticed that there are some `textField` entries in the `imageFields` array in this template, might be worth removing those in this PR as well

jacobkeeler (approved):

Issue #182, opened 2019-07-23
Media template bugs


NicoleYarroch:
1. If app's `appHMIType` is **not** `MEDIA` attempting to set the media template using `SetDisplayLayout` should fail. Currently it succeeds.
2. If app's `appHMIType` is **not** `MEDIA`, the `MEDIA` template should **not** be in `templatesAvailable` returned by the RAI response. Currently, the `MEDIA` template is returned even when the app type is non-media (i.e. `COMMUNICATION`)

joeygrover:
@JackLivio can you look at this?

iCollin:
I do not agree with #2. The capabilities in RAI response are NOT app specific and therefore do not depend on an app's HMI type.

A fix for #1 is available in https://github.com/smartdevicelink/generic_hmi/pull/419

iCollin:
Closed via https://github.com/smartdevicelink/generic_hmi/pull/419

PR #419, opened 2021-08-11
remove MEDIA template from non-media app capabilities

iCollin:
Fixes https://github.com/smartdevicelink/generic_hmi/issues/182

### Summary
When sending OnSystemCapabilityUpdated, only include MEDIA template in templatesAvailable if an app is media type.
When receiving Show request, do not allow non-media app to use template MEDIA.

SetDisplayLayout is unmodified as it is deprecated.

The issue also incorrectly requests changes to the RAI response, the capabilities in the RAI response are not App specific and should not be influenced by the app type. The capabilities in the RAI response are unchanged by this PR.

### CLA

### DIFF
--- a/src/js/Controllers/RpcFactory.js
+++ b/src/js/Controllers/RpcFactory.js
@@ -230,8 +230,8 @@ class RpcFactory {
         return msg
     }
     static UIGetCapabilitiesResponse(rpc) {
-        var displayCapabilities = capabilities["MEDIA"].displayCapabilities;
-        displayCapabilities.templatesAvailable.push('MEDIA');
+        let displayCapabilities = { ...capabilities["MEDIA"].displayCapabilities };
+        displayCapabilities.templatesAvailable = [...displayCapabilities.templatesAvailable, 'MEDIA'];
         return ({
             "jsonrpc": "2.0",
             "id": rpc.id,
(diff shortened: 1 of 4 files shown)


src/js/Controllers/UIController.js
```diff
@@ -91,16 +92,21 @@ class UIController {
                     // Generic HMI only supports main window for now.
                     return false;
                 }
+                var showApp = store.getState().appList.find((app) => {
+                    return app.appID === rpc.params.appID;
+                });
+                var showAppIsMedia = showApp.appType.includes('MEDIA');
```
JackLivio, line 98: HMIApplicationStruct has an additional (redundant) parameter to identify an app type. `isMediaApplication`
```suggestion
                var showAppIsMedia = showApp.appType.includes('MEDIA') || showApp.isMediaApplication;
```

untested ^

iCollin: This param is only in the `OnAppRegistered` notification and the app list state is only updated by the `UpdateAppList` notification, the `isMediaApplication ` value is not stored anywhere in the state.

I thought this value was redundant as well, do you think it is worth adding this to the state? It wouldn't be straight forward to do because the OnAppRegistered arrives before UpdateAppList.

JackLivio: I am seeing isMediaApplication sent in the UpdateAppList rpc
Screen Shot 2021-08-12 at 2 21 40 PM

src/js/Controllers/RpcFactory.js
```diff
@@ -311,7 +314,9 @@ class RpcFactory {
                 "id": rpc.id,
                 "error": {
                     "code": 1,
-                    "message": "The requested layout is not supported on this HMI",
+                    "message": (templateConfiguration.template === 'MEDIA'
+                        ? 'Only MEDIA apps may use the MEDIA template'
```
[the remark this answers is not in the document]
iCollin: Added in https://github.com/smartdevicelink/generic_hmi/pull/419/commits/0bcbcf1456effdc9fc18467f7f8134beac13abe6

src/js/Controllers/UIController.js
iCollin: Ah I was mistaken about that. That is the proper value to check, I have changed the appType checks in https://github.com/smartdevicelink/generic_hmi/pull/419/commits/0bcbcf1456effdc9fc18467f7f8134beac13abe6

JackLivio (approved):

Issue #296, opened 2020-09-17
The text in mainField1 overflows in the Non-Media template if the text is too long


NicoleYarroch:
## Bug Report
If the text set for `mainField1` is too long, it pushes down additional text and the text set in `mainField3` behind the soft buttons. 

### Reproduce
1. Send a `Show` request with the template set to `NON-MEDIA` and `mainField1` to a long string. 

#### Expected Result
1. The text in `mainField3` should take up at least 1 line and the remaining text should be truncated with ellipsis.
1. The text in `mainField1` should fill the rest of the available space and the remaining text should be truncated with ellipsis.

#### Actual Result
The additional text in `mainField1` is visible behind the soft buttons and the text in `mainField3` is not visible. 

<img width="963" alt="mainField1" src="https://user-images.githubusercontent.com/19915981/93513188-81676600-f8f3-11ea-8e44-dcf9d935c21b.png">

iCollin:
Closed via https://github.com/smartdevicelink/generic_hmi/pull/425

PR #425, opened 2021-08-17
Restrict overflow of main fields in non media template

iCollin:
Fixes https://github.com/smartdevicelink/generic_hmi/issues/296

### Testing Plan

### Summary
Add CSS to restrict the size of the main fields so that they cannot be overflowed.

### CLA

### DIFF
--- a/src/js/Templates/NonMedia/NonMediaBody.js
+++ b/src/js/Templates/NonMedia/NonMediaBody.js
@@ -18,14 +18,14 @@ export default class NonMediaBody extends React.Component {
         if(textFields.length > 2) {
             for(i=0; i<textFields.length; i++) {
                 if(i < 2) {
-                    textFields[i].className = "t-large t-light th-f-color non-media-text-4"; 
+                    textFields[i].className = "t-large t-light th-f-color non-media-text-4 non-media-oneline";
                 } else {
                     textFields[i].className = "t-small t-light th-f-color-secondary non-media-text-4"; 
                 }
             }
         } else {
             if(textFields[0]) {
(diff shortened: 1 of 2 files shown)


src/js/Templates/NonMedia/NonMediaBody.js
```diff
@@ -18,14 +18,14 @@ export default class NonMediaBody extends React.Component {
         if(textFields.length > 2) {
             for(i=0; i<textFields.length; i++) {
                 if(i < 2) {
-                    textFields[i].className = "t-large t-light th-f-color non-media-text-4"; 
+                    textFields[i].className = "t-large t-light th-f-color non-media-text-4 non-media-oneline";
                 } else {
                     textFields[i].className = "t-small t-light th-f-color-secondary non-media-text-4";
```
ShobhitAd, line 23: Probably should also add the overflow changes for mainField3 and mainField4
Screenshot from 2021-08-18 14-41-50

ShobhitAd (changes requested):

src/js/Templates/NonMedia/NonMediaBody.js
iCollin: Good catch, fixed with the `word-break: break-all;` property in https://github.com/smartdevicelink/generic_hmi/pull/425/commits/967fce9aa745ad906c444b53862a68de124ff383

ShobhitAd: Nice, that seems to have fixed the issue

ShobhitAd (approved):

PR #383, opened 2021-04-19
Update supported platforms in README.md

jacobkeeler:
### Testing Plan
- [x] I have verified that I have not introduced new warnings in this PR (or explain why below)
- [x] I have tested this PR against Core and verified behavior (if applicable, if not applicable, explain why below).

Core version / branch / commit hash / module tested against: 7.1.0
Proxy+Test App name / version / branch / commit hash / module tested against: N/A

### Summary
Updates supported platforms in README.md

##### Enhancements
* Updates supported platforms in README.md

### CLA
- [x] I have signed [the CLA](https://docs.google.com/forms/d/e/1FAIpQLSdsgJY33VByaX482zHzi-xUm49JNnmuJOyAM6uegPQ2LXYVfA/viewform)

PR #454, opened 2021-09-24
Update supported platforms in README.md (#383)

iCollin:
backmerge hotfix from last release


iCollin:
We need to remove 16.04 from supported platforms, so it is not necessary to back merge this commit

Issue #453, opened 2021-09-24
HMI does not highlight soft buttons on media screen


GetmanetsIrina:
### Bug Report
HMI does not highlight soft buttons on the media screen

##### Reproduction Steps
1. HMI and SDL are started
2. Media mobile app is registered and activated
3. Mobile app requests Show with softButtons(type = "IMAGE", isHighlighted = true)

##### Expected Behavior
HMI highlights added buttons

##### Observed Behavior
HMI does not highlight `softButtons` on the media screen

##### Browser & Version Information
* Browser and Version: Google Chrome (Version 92.0.4515.159)
* Generic HMI Version: release/0.11.0 (838d06e3)
* Testing Against: SDL: release/8.0.0 (de58a410) , SyncProxyTester(Version 20210729-Android (1082))

GetmanetsIrina:
Please notice Luxoft team is going to prepare a fix.

iCollin:
Closed via https://github.com/smartdevicelink/generic_hmi/pull/455

Issue #135, opened 2018-11-21
Add the ability to highlight soft buttons


NicoleYarroch:
Add the ability to highlight soft buttons. This can come in useful if developers want to have a selected and unselected state for a soft button as with the addition of the templated image feature we can no longer use icon color as an indicator of state change.

NicoleYarroch:
@JackLivio

iCollin:
Closed via https://github.com/smartdevicelink/generic_hmi/pull/426

PR #455, opened 2021-09-27
Fix HMI does not highlight soft buttons on media screen

ValeriiMalkov:
Fixes #[[453](https://github.com/smartdevicelink/generic_hmi/issues/453)]

### Testing Plan

### Summary
Added `backgroundColor` for static icon

### CLA

### DIFF
--- a/src/js/Templates/Shared/StaticIcon.js
+++ b/src/js/Templates/Shared/StaticIcon.js
@@ -13,7 +13,7 @@ export default class StaticIcon extends React.Component {
             var image = hex + value;
             path = "./" + image + ".svg"
             return (
-                <div className={this.props.class} >
+                <div className={this.props.class} style={{backgroundColor: this.props.fillColor}} >
                     {iconCache[path]}
                 </div>
             )
--- a/src/js/containers/Buttons.js
+++ b/src/js/containers/Buttons.js
@@ -29,7 +29,8 @@ const mapStateToProps = (state) => {
             image: softButtons[0].image ? softButtons[0].image.value : undefined,
             imageType: softButtons[0].image ? softButtons[0].image.imageType : undefined,
             id: softButtons[0].softButtonID,
-            isTemplate: softButtons[0].image ? softButtons[0].image.isTemplate : null
+            isTemplate: softButtons[0].image ? softButtons[0].image.isTemplate : null,
+            isHighlighted: softButtons[0].isHighlighted
         })
     }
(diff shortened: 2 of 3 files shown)


ValeriiMalkov:
Hi @theresalech ! This PR relates to the issue [#135](https://github.com/smartdevicelink/generic_hmi/issues/135) which is included in scope 0.11.0

iCollin:
Thank you for submitting this fix. As we are now finalizing our testing of Generic HMI 0.11, we are only focusing on regression issues. We will review this PR when time allows after the 0.11 release.

src/js/containers/Buttons.js
```diff
@@ -29,7 +29,8 @@ const mapStateToProps = (state) => {
             image: softButtons[0].image ? softButtons[0].image.value : undefined,
             imageType: softButtons[0].image ? softButtons[0].image.imageType : undefined,
             id: softButtons[0].softButtonID,
-            isTemplate: softButtons[0].image ? softButtons[0].image.isTemplate : null
+            isTemplate: softButtons[0].image ? softButtons[0].image.isTemplate : null,
+            isHighlighted: softButtons[0].isHighlighted
```
iCollin, line 33: a similar change needs to be applied to the second controlbaritem added around line 150

src/js/Templates/Shared/StaticIcon.js
```diff
@@ -13,7 +13,7 @@ export default class StaticIcon extends React.Component {
             var image = hex + value;
             path = "./" + image + ".svg"
             return (
-                <div className={this.props.class} >
+                <div className={this.props.class} style={{backgroundColor: this.props.fillColor}} >
```
iCollin, line 16: this background color should be applied in the control bar item class to the parent of the static icon, not in this class, as the bar item (soft button) is what is highlighted.

with the background, we should at least add `border-radius: 5px;` as well

src/js/containers/Buttons.js
ValeriiMalkov: @iCollin Applied in 351e590

src/js/Templates/Shared/StaticIcon.js
ValeriiMalkov: @iCollin Please take a look changes in commit 351e590
The `border-radius` added to `_control-bar.scss`. `backgroundColor` added as a `style` property of a `div` in a `ControlBarItem.js`

Issue #463, opened 2021-10-08
HMI deploy_server.sh is unable to process arguments


VjKlepikov:
### Bug Report
HMI deploy_server.sh is unable to process arguments

- cert-passphrase
- host
- ws-port
- video-port
- audio-port
- fs-port
- fs-uri

##### Reproduction Steps
1. Execute ./deploy_server.sh – cert-passphrase <CERTIFICATE_PASSPHRASE>

##### Expected Behavior
deploy_server.sh successfully process argument

##### Observed Behavior
deploy_server.sh does not process argument

##### Browser & Version Information
SDL: https://github.com/smartdevicelink/sdl_core/tree/release/8.0.0 (de58a41)
Generic HMI: https://github.com/smartdevicelink/generic_hmi/tree/release/0.11.0 (ec5639b)
SPT: 20210729-Android(1082)
PolicyServer: Version: 2.11.0 https://github.com/smartdevicelink/sdl_server/tree/develop (4a804c6)

PR #465, opened 2021-10-08
Fix/deploy server unable to process arguments

ValeriiMalkov:
Fixes [#463 ](https://github.com/smartdevicelink/generic_hmi/issues/463)

### Testing Plan

Core version https://github.com/smartdevicelink/sdl_core/tree/8.0.0-RC1 (86e032f)
HMI: https://github.com/smartdevicelink/sdl_hmi/tree/release/5.6.0 (1fb8b6b)
Mobile: SPT 20210729-Android, Proxy Lib Ver: 2020-11-04

### Summary
Added arguments proxing to the python script from bash script

### CLA

### DIFF
--- a/tools/start_server.py
+++ b/tools/start_server.py
@@ -50,6 +50,8 @@ import OpenSSL.crypto as crypto
 
 class Flags():
   """Used to define global properties"""
+  HOST = '127.0.0.1'
+  WS_PORT = 8081
   FILE_SERVER_HOST = '127.0.0.1'
   FILE_SERVER_PORT = 4000
   FILE_SERVER_URI = 'http://127.0.0.1:4000'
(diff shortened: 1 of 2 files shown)


tools/start_server.py
```diff
@@ -50,6 +50,8 @@
 
 class Flags():
   """Used to define global properties"""
+  HOST = '127.0.0.1'
```
iCollin, line 53: since these aren't global properties and used in other places like all other flags, i don't think these should be added to this class. I think we can just put the values in the default= expression without declaring them elsewhere.

iCollin (changes requested):
Bash has argument forwarding already built in, instead of implementing that ourselves, check out this patch for the bash changes:

```
diff --git a/deploy_server.sh b/deploy_server.sh
index 1baaeafc..717ea4d9 100755
--- a/deploy_server.sh
+++ b/deploy_server.sh
@@ -39,7 +39,7 @@ InitSubmodules() {
 
 StartServer() {
     cp ${SOURCE_DIR}/${TARGET_SCRIPT} ${TARGET_DIR}
-    python3 ${TARGET_DIR}/${TARGET_SCRIPT} --host 127.0.0.1 --ws-port 8081
+    python3 ${TARGET_DIR}/${TARGET_SCRIPT} "$@"
     rm ${TARGET_DIR}/${TARGET_SCRIPT}
 }
 
@@ -49,6 +49,6 @@ if ! find $TARGET_DIR -mindepth 1 | read; then
 fi
 
 echo "Starting HMI Backend service..."
-StartServer
+StartServer "$@"
 echo "HMI Backend service was stopped"
```

tools/start_server.py
ValeriiMalkov: ADded in commit 30ac327

PR #476, opened 2021-10-29
Fix/Get default layout based on app type

### LINKED ISSUE #448: Setting DEFAULT layout does not take appHMIType into account
### Bug Report
When setting the template to `DEFAULT`, the Generic HMI will always set the layout to the `MEDIA` layout (even though non-media apps normally don't have access to this template). The HMI should take the appHMIType of the app into account when setting the layout to `DEFAULT`, as is done when the app is first registered

##### Reproduction Steps
1. Register app with `INFORMATION` appHMIType and activate app
2. Send a Show message with `templateConfiguration.template` set to `DEFAULT`

##### Expected Behavior
`NON-MEDIA` template is still displayed

##### Observed Behavior
`MEDIA` template is displayed

##### Browser & Version Information
* Browser and Version: Google Chrome Version 92.0.4515.131 (Official Build) (64-bit)
* RPC Sent: 
```
{
  "mainField1":"MAIN1",
  "mainField2":"MAIN2",
  "graphic":{
    "value":"0x29",
    "imageType":"STATIC"
  },
  "softButtons":[
    {
      "type":"BOTH",
      "softButtonID":1,
      "text":"ONE",
      "image":{
        "value":"0xD7",
        "imageType":"STATIC"
      }
    },
    {
      "type":"BOTH",
      "softButtonID":2,
      "text":"TWO",
      "image":{
        "value":"0xD8",
        "imageType":"STATIC"
      }
    }
  ],
  "templateConfiguration":{
    "template":"DEFAULT"
  }
}
```
* Output Received: N/A (success response)
* Generic HMI Version: `release/0.11.0`
* Testing Against: RPC Builder App JS

ShobhitAd:
Fixes #448 

### Testing Plan

#### Core Tests
[List of tests performed against Core and behaviors verified]

Core version / branch / commit hash / module tested against: [INSERT]
Proxy+Test App name / version / branch / commit hash / module tested against: [INSERT]

### Summary
[Summary of PR changes]

### CLA

### DIFF
--- a/src/js/Controllers/UIController.js
+++ b/src/js/Controllers/UIController.js
@@ -280,11 +280,12 @@ class UIController {
                 if (disallowedLayout) {
                     rpc.params.displayLayout = prevDisplayLayout;
                 }
-                let newTemplate = (rpc.params.displayLayout === 'DEFAULT') ? 
-                    getDefaultLayout(rpc.params.appID, setDisplayLayoutApp) : rpc.params.displayLayout;
-                store.dispatch(setTemplateConfiguration(newTemplate, rpc.params.appID, rpc.params.dayColorScheme, rpc.params.nightColorScheme));
+                if (rpc.params.displayLayout === 'DEFAULT') {
+                    rpc.params.displayLayout = getDefaultLayout(rpc.params.appID, setDisplayLayoutApp)
+                }
+                store.dispatch(setTemplateConfiguration(rpc.params.displayLayout, rpc.params.appID, rpc.params.dayColorScheme, rpc.params.nightColorScheme));
 
-                if (prevDisplayLayout !== newTemplate) {
+                if (prevDisplayLayout !== rpc.params.displayLayout) {
                     this.listener.send(RpcFactory.OnSystemCapabilityDisplay(rpc.params.displayLayout, rpc.params.appID));
                 }
                 return {"rpc": RpcFactory.SetDisplayLayoutResponse(rpc, disallowedLayout)};
--- a/src/js/Controllers/UIController.js
+++ b/src/js/Controllers/UIController.js
@@ -127,7 +127,7 @@ class UIController {
                         templateConfiguration.nightColorScheme
                     ));
                     
-                    if (prevDisplayLayout !== templateConfiguration.template) {
+                    if (prevDisplayLayout !== appUIState.displayLayout) {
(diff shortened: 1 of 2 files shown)


jacobkeeler (changes requested):

ShobhitAd:
Found another bug related to `OnSystemCapabilityUpdated(DISPLAYS)` being sent every time the display layout is changed (even if the requested layout is the same as the current layout).

It looks like the `prevDisplayLayout !== newTemlate` checks, for both Show and SetDisplayLayout, are invalid since `prevDisplayLayout` is in the lower case format ('media', 'nonmedia', 'text-and-softbuttons-with-graphic', etc.) while the `new Template` variable is in the regular template name format ('MEDIA', 'NON-MEDIA', `TEXT_AND_SOFTBUTTONS_WITH_GRAPHIC`, etc.) so the comparison would fail every time.

Fixed the issue in https://github.com/smartdevicelink/generic_hmi/pull/476/commits/d24885b3685db8580b2c5a8b9cb910ee9fc534e6 by using `prevDisplayLayout !== appUIState.displayLayout` instead of the original comparision

src/js/Controllers/UIController.js
```diff
@@ -262,10 +275,12 @@ class UIController {
                 if (disallowedLayout) {
                     rpc.params.displayLayout = prevDisplayLayout;
                 }
-
+                if (rpc.params.displayLayout === 'DEFAULT') {
+                    rpc.params.displayLayout = getDefaultLayout(setDisplayLayoutApp)
+                }
                 store.dispatch(setTemplateConfiguration(rpc.params.displayLayout, rpc.params.appID, rpc.params.dayColorScheme, rpc.params.nightColorScheme));
```
jacobkeeler, line 281: I didn't realize that this event was guaranteed to finish before moving on in the thread (I always assumed this was asynchronous), but after repeated tests it looks like that is the case. The OnSystemCapabilityUpdated message is always sent in the correct situations from what I can tell.

jacobkeeler (approved):

src/js/Controllers/UIController.js
ShobhitAd: That was my initial assumption as well but looking at https://redux.js.org/api/store#dispatchaction it looks like dispatch is always synchronous

PR #504, opened 2022-03-23
Add custom masking font

JackLivio:
Fixes a regression issue in 0.12.0

### Testing Plan

#### Core Tests
Send setGlobalProperties
```
{ "keyboardProperties": { "maskInputCharacters": "USER_CHOICE_INPUT_KEY_MASK" } }
```

Display perform interaction keyboard. Confirm text input is hidden in chrome and firefox.

Core version / branch / commit hash / module tested against: 8.1
Proxy+Test App name / version / branch / commit hash / module tested against: rpcb-js

### Summary
Firefox does not support the css param

```
-webkit-text-security: disc;
```

Therefore I added a font that hides the text input on the font level to support font hiding in firefox.

### CLA


src/css/components/_keyboard.scss
```diff
@@ -203,5 +203,6 @@
 }
 
 .text-security {
+    font-family: "LivioMask";
     -webkit-text-security: disc;
 }
```
ShobhitAd, line 208: The size of the masked characters is somewhat larger in firefox. It might be better to make the following change to make the masked characters consistent on both chromium and firefox
```suggestion
.text-security {
    font-family: "LivioMask";
    font-size: 18px;
}
```

ShobhitAd (changes requested):

ShobhitAd (approved):

